PR #29, opened 2014-12-29
Fix: dropdown menu on click

eri-trabiccolo:
This is a better version of my previous pull-request.
I discovered it didn't work when a first level menu item was a parent and a link to a content.
Was this the reason you changed it?
Also consider that the check on href!='' is useless since we force all parent menu-items to have
href="#"


eri-trabiccolo:
Nicolas,
I've made some fixes to this.
Cheers

PR #30, opened 2014-12-31
Add new filters for fp and footer widget areas

eri-trabiccolo:
I think these two filters could be useful, for example when people change the .container width
You might want to apply the fp one to fpu also.
Bests

edit:
I see you already have something of the sort for fpu :D


eri-trabiccolo:
Ops.. didn't know str_replace works also with array :)
forget my comment above :)

PR #31, opened 2015-01-01
Fix logo print: don't use wrong attachment's height & width

eri-trabiccolo:
Looks like that when you upload an svg (or at least the one used here:
https://wordpress.org/support/topic/svg-images-not-working
 )
wp_get_attachment_src(...) may return wrong size values (1,1).

Could just this be a correct fix?


Nikeo:
Good point.
I 'll test this and let you know.
thanks

Nikeo:
Merged in dev thanks

PR #32, opened 2015-01-04
Fix: retrieve correct ID in posts page

eri-trabiccolo:
see: http://themesandco.com/snippet/display-post-metas-home-page/comment-page-1/#comment-289929
when wp_query->is_posts_page == true, and in the loop, basically get_query_object()->ID returns always the id of the first post.

Still regarding post metas:
https://wordpress.org/support/topic/eliminate-about-at-end-of-each-blog-page
this is about conditional options
I see this function (slightly different though) it's also in some extensions.
Suggestion:
if extensions are running on customizr(-pro) don't add filters already added in the theme.

### DIFF
--- a/inc/class-fire-utils.php
+++ b/inc/class-fire-utils.php
@@ -251,7 +251,7 @@ if ( ! class_exists( 'TC_utils' ) ) :
       */
       function tc_get_the_ID()  {
           global $wp_version;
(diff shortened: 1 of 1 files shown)


Nikeo:
Good point.
Thanks
I'll test and merge this

Nikeo:
Merged in dev

giorgioriccardi:
great! I'll test it as well as soon as it is available

PR #33, opened 2015-01-11
Fix donate and cta in subpanel

eri-trabiccolo:
Hide #tc-donate-customizer in subpanels 
Add transition effects both to #tc-donate-customizer and .tc-cta-wrap


Nikeo:
coool

PR #34, opened 2015-01-11
Fix thumb inline style

eri-trabiccolo:
Pretty invisible typo, 'cause it's overridden by js centering images.
;)
Gotta go now, talk you later.

rock


Nikeo:
yes hard to detect!
Fixed manually directly into dev.

PR #35, opened 2015-01-19
Fix: check if wp_query var exists

eri-trabiccolo:
otherwise a notice will be displayed (plug-in not running) when WP_DEBUG=true


Nikeo:
Merged thanks !

PR #38, opened 2015-01-23
Fix: don't re-add edit link button for tribe-events events

eri-trabiccolo:
use tribe-events api to determine whether or not to perform compatibility actions


eri-trabiccolo:
Basically we don't need to re-add that link button for two reasons:
1) In tribe events post lists that edit link is hidden with css by default. So nobody cares about it, and nobody expects to have it
2) In tribe events single posts the edit link isn't printed anyway, since the condition in_the_loop() ( TC_headings->tc_add_edit_link_after_title() ) fails.

eri-trabiccolo:
Wonderful why travis says it has an error?
In my travis it passed.. :/
Anyway I saw that travis detected a parse error when you put that code in dev: https://travis-ci.org/Nikeo/customizr/jobs/47553965#L301
But didn't really warn you.. maybe depends on some settings, who knows..

eri-trabiccolo:
:)

PR #39, opened 2015-01-24
Fix: dropcap, do nothing if the first p isn't valid

eri-trabiccolo:
e.g. : if the first paragraph is something like <p></img></p>, _first_p_text is null
therefore _first_p_text.replace() throws Uncaught TypeError


eri-trabiccolo:
There could be other unhandled cases.

eri-trabiccolo:
Though would be better to move on the following paragraph till you find a valid one. Think about when you have an image as first p and a regular text as second. People might expect the dropcap working on what they think is the first p. Also, if the image has alignright, they really wouldn't understand why dropcap isn't applied..

PR #41, opened 2015-01-24
Fix: force hiding #tc-reset-margin-top when no sticky header

eri-trabiccolo:
A possible solution to the following problem:
Sticky header enabled.
I see that opening the mobile menu makes the header not fixed (removing the body class "tc-sticky-header"), but #tc-reset-margin-top is still shown (and its margin valued). This is only tolerable when you have a menu with an height >= of the #tc-reset-margin-top's margin (header's height).


Nikeo:
merged in dev

PR #42, opened 2015-01-24
Imp: handle .tc-no-title-logo, when sticky enabled, in css

eri-trabiccolo:
Tested, but you better test it too :P

### DIFF
--- a/inc/assets/js/parts/main.js
+++ b/inc/assets/js/parts/main.js
@@ -484,15 +484,6 @@ jQuery(function ($) {
         $tcHeader.css('top' , _get_initial_offset() + 'px');
     }
 
-    function _set_no_title_logo_class() {
-        if ( ! $('body').hasClass('sticky-enabled') ) {
-            $('.navbar-wrapper').addClass('span9').removeClass('span12');
-        } else {
-            $('.tc-title-logo-off .navbar-wrapper' , '.sticky-enabled').addClass('span12').removeClass('span9');
-            $('.tc-title-logo-on .navbar-wrapper' , '.sticky-enabled').addClass('span9').removeClass('span12');
-        }
-    }
-
     //set site logo width and height if exists
     //=> allow the CSS3 transition to be enabled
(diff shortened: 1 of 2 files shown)


Nikeo:
Hi Rocco, can you recall me what is the problem solved here? Is it the issue reported for a big logo ?

eri-trabiccolo:
Nope.. Is the issue about that little flashing an user on the pro support () reported. Basically there's a little range in which sticky is enabled but navbar-wrapper is still span9. Then scrolling up it become 12 and then 9 again. This fixes the issue and is also very simple.

eri-trabiccolo:
Ah, also, using span9, statically, isn't very safe, I think, since through filters, you can change that value, so if you still want to use the js method (I disagree, but it's your theme :P) you should pass it in TCParams, like you do with the sidebars...
Though it will still work when you use spanX with X < 9, since, in the css, they come after 9, so span9 css is overridden, but if a user changes the navbar-wrapper class to span10, then it's wrong..

Nikeo:
I am convinced by your CSS solution.
To be merged ! thanks

Issues : yes I will!

Nikeo:
merged in dev

Nikeo:
About issues : I don't think there are github issues to adress ?
Are you mentioning the forum issues on niclook ?

eri-trabiccolo:
Nope, I use niclookbug for those :P 
I use nicklook to inform you of good things, or to let you know how greatly I handled a certain situation :P 
Eheh
I meant the extensions' issues (wfc/fpu).
About customizr issues, the missing script is solved right? So I can close the last two I opened

Nikeo:
ah ah!
OK yes the wfc/fpu.
Yes missing script solved.

I'm done for today ! good evening Rocco.

eri-trabiccolo:
Good evening too :)

eri-trabiccolo:
You're served, now there's a new entry in niclookbug :P

PR #43, opened 2015-01-25
Fix: handle bool value in default options

eri-trabiccolo:
See issue https://github.com/Nikeo/customizr/issues/40


eri-trabiccolo:
Well, this is partial, while I was playing football I thought that should handle also type==select. But I need some rest now :)

Nikeo:
merged into dev

eri-trabiccolo:
Why I see this closed and not merged? Weird..
Did you manually modify the file? (just to know how github works :D)

PR #44, opened 2015-01-28
Fix: in post metas don't call tc_has_update() when not needed

eri-trabiccolo:


eri-trabiccolo:
Though a valid pull request, 'cause if the user don't want to display the update notice, there's no sense in calling it at that point, this is more a reminder for the possible bug reported here:
https://wordpress.org/support/topic/posts-page-posts-not-showing-in-another-language?replies=3

eri-trabiccolo:
Mmm travis says it has an error, but looks like an issue with a not installed php version!?
Anyway I handled a basic try, catch block. This should be fine, but we should go to the root of the issue (and at the moment I don't know how to reproduce it) now I just say that if some error occurs you don't have to show the update.

PR #45, opened 2015-01-28
Fix: don't print menu item first letter style by default

eri-trabiccolo:
Customizr doesn't differently style the first letter of a menu item anymore
[reminder]


Nikeo:
Updated in dev thanks :)

eri-trabiccolo:
Not updated for Customizr-Pro: TC_utils::$instance -> tc_user_started_before_version( '3.3.0' )
This is something you should to consider every time you use tc_user_started_before_version
so, for example, in the latest updates , for the :before icons.

Nikeo:
oh yes, very good point !

PR #46, opened 2015-01-29
zh_TW.po

pppdog:
Add lang/zh_TW.po


Nikeo:
added to dev.
Thanks for your contribution :)

PR #48, opened 2015-01-31
Post list design

eri-trabiccolo:
Ok, I left some commented parts here and there, to let you decide ;)
Opacify the thumb + white font color, fine but we'll have issues with thumb with predominant white.


eri-trabiccolo:
Would be good to squash all these in a single commit.

Nikeo:
yes but that's not a big deal.
I don't how to do this anyway :)
I have pulled your pulled request into a local branch.
I am currently working on it. I'll keep you posted

eri-trabiccolo:
Since that's no a Customizr issue, I suggest you to ask on the git forum :P ahahah
something interesting here: http://makandracards.com/makandra/527-squash-several-git-commits-into-a-single-commit

Anyway about the feature, at first glance, do you think it's somewhat satisfactory or there is a lot to trash/improve?

Nikeo:
that's just fine !

PR #49, opened 2015-02-02
Fix: don't escape html entities in slide-text but strip tags

eri-trabiccolo:
https://wordpress.org/support/topic/text-on-slider-getting-truncated-even-though?replies=3
This is because "'" and "&" (in that case) are saved, as "&quot;" and "&amp;" .
I make this pull request but I'm not sure this is THE solution, I think you can find a better one.
So, take this as a reminder ;)


eri-trabiccolo:
About this, I realized that could be not compliant with theme review guidelines . If so, stash this, and we'll just suggest to increase the char limit for the slider caption.

PR #50, opened 2015-02-04
Fix: fix posts with no title preventing displaying archive titles

eri-trabiccolo:


eri-trabiccolo:
I forgot this, IMHO in 3.3.0 too

eri-trabiccolo:
Not needed anymore since https://github.com/Nikeo/customizr/commit/bd9c37003a44c598884cf450d801a7f2d6789822

PR #51, opened 2015-02-06
Author URI in style.css generates wrong link

giorgioriccardi:
This is actually for Customizr-Pro

Author URI: http://themesandco.com/contact/ in style.css Customizr-pro. This /contact/ generates a link error in /wp-admin/themes.php?page=welcome.php for the showcase button.

The final outcome is a wrong link: http://themesandco.com/contact/customizr/showcase/ in customizr free version it does not happen bec the URI is just http://themesandco.com/


giorgioriccardi:
I'm not sure what will happen with this pull-request, since it is for customizr-pro.
Let's see.
Hope it helps

giorgioriccardi:
Great you fixed it in customizr-pro 1.0.11 :-) glad it helpd

Nikeo:
Thanks Giorgio !

PR #52, opened 2015-02-07
Fix: add body class sticky-disabled, with php

eri-trabiccolo:
by default when sticky header selected.


eri-trabiccolo:
Fixes a small issue.
When you have selected logo centered the header is displayed, at first, as logo-left / navbar-right. Then on document ready logo and navbar are centered because of css rules depending on the body class sticky-disabled.

PR #55, opened 2015-02-07
Fix: new filter tc_the_title

eri-trabiccolo:
Apply filters to the_title in Customizr contexts strictly.

side notes:
Remove the_events_calendar compatibility functions useless now.
Keep the_events_calendar theme support, since it's still supported
Make woocommerce products still use the Customizr title

Using this new special filter means that Customizr title isn't used anywhere, this avoids plugin compatibility issues (with many plugins), but, as for woocommerce, implies that we have to specifically force some plugins to use the Customizr title, which is always better than write N patches IMHO :)

### DIFF
--- a/inc/class-fire-init.php
+++ b/inc/class-fire-init.php
@@ -717,7 +717,8 @@ if ( ! class_exists( 'TC_init' ) ) :
           }
 
         }//end if woocommerce
-      }
+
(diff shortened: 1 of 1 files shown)


Nikeo:
perfect :)

eri-trabiccolo:
Finally :P

PR #58, opened 2015-02-09
Fix: logo image stretched when sticky header enabled,

eri-trabiccolo:
handled for both logos (normal and sticky) in js.
Avoid using css 'height:auto !important', to keep transitions effects


eri-trabiccolo:
wait, you use the _refresh ..  which calls a resize, then you have an infinite loop. That's why I added the settimeout and not _refresh ... am I wrong?

Nikeo:
oops yes.
Fixed thanks

PR #59, opened 2015-02-09
Fix: Remove icons by default for new users,

eri-trabiccolo:
but keep icons for Archive and Widget titles.
Archive icons are pretty good.
Widgets without icons are too much anonymous


Nikeo:
This has to be done with a check of the version when the user started. (some users may like those icons... :) )
I'll handled it thanks

eri-trabiccolo:
Right, forgot about the fact that if not changed default options are taken and that I didn't have to change default options for old users. sorry

PR #61, opened 2015-02-09
Add: new option to wrap the header elements into a bootstrap container

eri-trabiccolo:


giorgioriccardi:
Thanks Rocco!
Funny enough this was an option that I was trying to "force" via css on the header.
Hopefully will be merged as well on the next customizr-pro version

eri-trabiccolo:
:) Yeah, dunno if Nico likes it .. we'll see, but if he doesn't you can always use it in your child theme functions.php

giorgioriccardi:
Let's see! Hopefully he will like it, I do.
I tried to implement the function into functions.php, but I was not able to work out the part in "class-fire-utils_settings_map.php", so I put your files in my child-theme.
I know is not wise and is not going to work on a long therm, but I'm messing up on this test site [http://elc.workingdesign.ca/] and I'm running out of time to deliver a demo. 
So I'll have to figure out how to enforce both your suggestions in functions.php in the future.

eri-trabiccolo:
Hi Giorgio, in your child-theme functions.php just put this code:
https://gist.github.com/eri-trabiccolo/c9ff7f701bec02be3be0

giorgioriccardi:
LOL! that was actually simpler than I thought. 
Many thanks, really much appreciated all your effort

eri-trabiccolo:
:)

PR #62, opened 2015-02-12
Fix: better check on when changing default slider's height

eri-trabiccolo:
Reported by @daymobrew
Slightly different version of his proposed patch.
ref: https://wordpress.org/support/topic/notice-undefined-index-tc_slider_default_height


eri-trabiccolo:
Don't you want include this one in the next release?

Nikeo:
mmm you're right, I'll include it in v3.3.0.

eri-trabiccolo:
Ok, probably nobody will notice it ;)

damiencarbery:
No need for me to submit my patch as this one is better. Thanks for fixing.

PR #64, opened 2015-02-14
Fix: dropcap skips also ul,ol tags

eri-trabiccolo:
Fixes this bug:
https://wordpress.org/support/topic/initial-with-bullet-points-on-customizr?replies=2#post-6568402


eri-trabiccolo:
.. but as you can see there's still an issue with paragraphs which start with "(Something"

Nikeo:
ok.
has to be added in tc-js-params.js. (tc-script is the result of a grunt js concat task)
Fixed in dev
Thanks

eri-trabiccolo:
Yeah, I forgot it this time... I went too fast, sorry

PR #65, opened 2015-02-15
Add: add new body class for the current layout

eri-trabiccolo:


eri-trabiccolo:
Feel free to trash it if you don't like the "idea".

Nikeo:
Why not but do you have an example of use case / situation when this particular layout class is missing ?

eri-trabiccolo:
I thought at it here -> http://themesandco.com/support-forums/topic/sidebar-widget-not-resizing-on-tablet/#post-50527

In this case I proposed to but left and right borders to the #content but it's right when you use both sidebars, so I thought that in general having a css class for the current layout would have been useful

Nikeo:
OK got the point thanks

Nikeo:
added in dev

PR #66, opened 2015-02-16
Fix: typo in dutch translation

eri-trabiccolo:
ref: http://themesandco.com/support-forums/topic/size-social-icons-in-sidebar/


eri-trabiccolo:
merged

PR #71, opened 2015-02-21
Fix: hide/don't show social icons in footer when option unchecked

eri-trabiccolo:
ref: https://wordpress.org/support/topic/social-links-in-footer-1


giorgioriccardi:
it says file changed 0.

eri-trabiccolo:
yeah weird... but it's merged: https://github.com/Nikeo/customizr/blob/dev/inc/parts/class-footer-footer_main.php#L128

giorgioriccardi:
no idea, but I can't pull it from sourcetree, I guess I'll dump all and start from scratch

PR #72, opened 2015-02-26
Fix: amend conflicting actions when dropdowntoViewport false

eri-trabiccolo:
and link to an anchor clicked.
Perform those actions just when clicking on the menu button (responsive mode)

Fixes:
https://wordpress.org/support/topic/accordion-8
and
the issue with the Changelog accordions in TC

### DIFF
--- a/inc/assets/js/parts/bootstrap.js
+++ b/inc/assets/js/parts/bootstrap.js
@@ -1651,6 +1651,9 @@ var TCParams = TCParams || {};
 
       //@tc adddon
       //give the revealed sub menu the height of the visible viewport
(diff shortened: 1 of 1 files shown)


eri-trabiccolo:
Hi Nico, is it normal that this isn't already in tc-scripts.js after your merge?
Shouldn't be automatically generated with grunt?

Nikeo:
Hi !
tc-scripts.js is automatically generated by grunt each time one of the script is changed. But for that the watch task has to be enabled, which was not the case before I pushed...

Should be fine in now :)
Thanks

eri-trabiccolo:
:)

PR #74, opened 2015-03-02
Fix: smoot scroll compatibility with UPME

eri-trabiccolo:
User Profiles Made Easy compatibility.
When clicking on a button like <a href="#edit" class="upme-button-alt upme-fire-editor">View Profile</a>
"Uncaught TypeError: Cannot read property 'top' of undefined" was thrown
ref: http://themesandco.com/support-forums/topic/customizr-theme-js-some-plugins-not-working-with-this-theme/#post-57993


Nikeo:
Hi ! Did you test it

eri-trabiccolo:
Nope, it's a premium plugin but I inspected the element..
cannot find the link now :/

PR #75, opened 2015-03-03
Fix: do not display resp menu button when sticky enabled and tc-menu-off

eri-trabiccolo:
fixes: https://wordpress.org/support/topic/responsive-nav-menu-behavior-broken?replies=2#post-6642813

### DIFF
--- a/inc/assets/less/tc_custom_responsive.less
+++ b/inc/assets/less/tc_custom_responsive.less
@@ -371,6 +371,10 @@
     position: relative;
     z-index: 1;
   }
+  /* do not display menu button when sticky enabled and tc-menu-off enabled */
+  .sticky-enabled .tc-menu-off .btn-navbar {
+    display: none;
+  }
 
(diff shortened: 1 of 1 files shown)


eri-trabiccolo:
Follow up that topic:
Alternative solution: @media (max-width: 979px){
.sticky-enabled .nav-collapse.collapse.in{display: block}
}
don't prevent displaying of the nav-collapse in mobiles even if that option (tc-menu-off when scrolling + sticky header) is enabled.
This could be a further option in Customize (different behavior in mobile and desktop). But, since there are already a lot of options, my opinion is that sooner or later will be the time to chose to include the options that will be more likely used, and cover the rest with snippets. Signed humble_rock :P

Nikeo:
Ok thanks for reporting this.
I agree about the options choices that have to be made. Contx will give us some oxygen (front page option for example will only appear in one context).

PR #76, opened 2015-03-04
Add: new post list grid filter, new class init-pro

eri-trabiccolo:


eri-trabiccolo:
I think you can close this either. Isn't it?

PR #77, opened 2015-03-04
Fix: array dereferencing fix for PHP<5.4.0

eri-trabiccolo:
some_function()[I] introduced in php 5.4.0
(... sorry, it's my code. did travis said anything about that? )

expanded_featured check reengineered a little bit


eri-trabiccolo:
well, travis fails on this one, but don't think is about the code itself, right?

PR #81, opened 2015-03-05
Fix: some minor changes to avoid weird behavior when loading page

eri-trabiccolo:
and sticky header set, also fix logos dimension not set (_.isNumber(num) returned false on width and height attr values => pretty important! )

on page loading:
a) CSS: set top: 0 when sticky disabled in CSS, it will changed in js if needed (see admin bar)
b) CSS: make tc-margin-top always visible for tc-sticky-header body class, this will reduce the jumping of the content while loading the page (before js takes effect)
c) JS: in LOADING ACTIONS call also  _scrolling_actions(). This because sometimes when you refresh (browser) an already scrolled page (or go to an anchor in a different page), for some reason (at least with chrome) the scroll event isn't fired => the header stays with its original size.

### DIFF
--- a/inc/assets/js/parts/main.js
+++ b/inc/assets/js/parts/main.js
@@ -540,9 +540,10 @@ jQuery(function ($) {
         _set_header_top_offset();
         //process scrolling actions
         if ( $(window).scrollTop() > triggerHeight ) {
-            $('body').addClass("sticky-enabled").removeClass("sticky-disabled");
+            if ( $('body').hasClass('sticky-disabled') )
+                $('body').addClass("sticky-enabled").removeClass("sticky-disabled");
         }
-        else {
+        else if ( $('body').hasClass('sticky-enabled') ){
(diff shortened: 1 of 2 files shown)


eri-trabiccolo:
With the last commit we should avoid side effects if any. Since when sticky header enabled the body has "sticky-disabled":
1) if $(window).scrollTop() > triggerHeight => add sticky-enabled class and remove sticky-disabled class => fine, that's what we want in those case when browsers doesn't fire the scroll event when loading the page
2) otherwise since body hasn't the class sticky-enabled, nothing will happen => fine

eri-trabiccolo:
which part? just the css part I hope, right?

Nikeo:
js part but no worries, it should be fine

eri-trabiccolo:
Mmm, but the js part was already in the core code ... the only missing part of this pr should have been the css part .. mmm

PR #82, opened 2015-03-06
Fix: apply excerpt and content filters in post lists

eri-trabiccolo:
Fixes this: http://themesandco.com/support-forums/topic/front-page-video-and-formatting-not-showing-when-full-content-selected/#post-58985

ref: http://codex.wordpress.org/Function_Reference/the_content#Alternative_Usage

About the_excerpt, should be applied also in the grid? I think so, then it requires a css rule to remove the margin bottom to the <p> in the grid. If do you want to make this last change too, also the addon should be modified. Let me know.


eri-trabiccolo:
Also consider, that if someone wants to make a template to show a list of posts the "wp" hook is too early (templates are loaded after), so I (strongly) suggest to use something like "wp_head" (tested) as action hook for the __loop hooks https://github.com/eri-trabiccolo/customizr/blob/dev/inc/parts/class-content-post_list.php#L24

giorgioriccardi:
I actually had the above mentioned problem with customizr 1.0.15 (I had to switch back to 1.0.12).
My custom post list template does not show anymore posts, but only page navigation... dunno if it is related though.

eri-trabiccolo:
Yes it is :/

eri-trabiccolo:
Giorgio, 
for the problem with the custom post lists follow this -> https://gist.github.com/eri-trabiccolo/14ade3c986e9f3086843#file-custom_post_lists_comp-php

Nikeo:
fixed in dev thanks !

PR #88, opened 2015-03-15
Fix: first step, load rtl css also if WP_LANG not defined

eri-trabiccolo:
but is_rtl() and load the parent rtl css even if we use a child without our rtl css version.

More a reminder than a real PR. is_rtl() could totally replace the check on WP_LANG, but I'm not sure about wp backward compatibility that's why I kept both.
Also those checks could be done "better" using an actual if-else statement instead of the ternary operator, but I kept your structure, since you really love it :P 
There are others things to the relatively to the css, but at the moment couldn't figure out how do you generate the rtl tc_common.css . Any clue? :D


Nikeo:
rtl tc_common.css => look in grunt-tasks-config/less.js and you'll figure it easily :)

eri-trabiccolo:
I see, so just the body direction and unicode-bidi changes in the rtl version, isn't it?
If so, there's a lot of work to do :P

Nikeo:
Yes !

PR #90, opened 2015-03-16
Fix: don't print the menu wrapper when no menu set

eri-trabiccolo:
You might don't want to accept the whole p-r, but limit the "not printing" just to the mobile button when no menu is set.


eri-trabiccolo:
Wow but this PR is a mess.
I'll do two separate PRs, why did I make this mess, meh :/

PR #91, opened 2015-03-16
Fix: CenterImages js avoid using ir8 reserved words

eri-trabiccolo:
Looks like "class" is a reserved word for ie8..!?
More info here: https://wordpress.org/support/topic/expected-identifier-line-3-char-5937-tc-scriptsminjsver339?replies=5#post-6696463
we'll see what JTS_IL will answer.


Nikeo:
fixed in dev

eri-trabiccolo:
Mmm, not the "stray commas" part ;)
still there

Nikeo:
done => was it generating errors ?

eri-trabiccolo:
Well, haven't tested since I couldn't find a ie8 :D But accordingly with what that guy said looks like it did..

PR #92, opened 2015-03-17
Fix: apply css to reply class just when inside #content div

eri-trabiccolo:
or maybe #main-wrapper .reply, or body .reply ?


Nikeo:
Ok merged in dev

PR #93, opened 2015-03-18
Fix: use user settings for the long date format

eri-trabiccolo:
This is a proposal,
why don't we use the user choice for the date format when == long?

https://wordpress.org/support/topic/error-with-the-italian-date-format-in-the-meta-tags?replies=5

Also passed two useful params to the filter tc_date_meta
and used already retrieved variabled lines 174-175


Nikeo:
excellent fixes thanks !
Changes made in dev

PR #97, opened 2015-03-20
Fix: comment-reply use default add-below param

eri-trabiccolo:
Look that this is the default wordpress add-below param. So you can either totally remove the add-below param, but since there's a filter I decided to leave it.

This works here, though in TC it doesn't. This 'cause the <article> comment in TC doesn't have the id attribute like id="comment-N", why?


Nikeo:
Cool :)

Merged in dev.
I'll check the TC problem thanks.

PR #98, opened 2015-03-20
Fix: add customiz-pro version check for post-list grid default value

eri-trabiccolo:
I think I set the correct version, since, from the pro changelog:
= 1.0.13 March 2nd 2015 =

    Upgraded to customizr v3.3.6


eri-trabiccolo:
Don't forget this, though if pro users have already updated their versions it's pretty useless :D

Nikeo:
Fixed in dev thanks.

PR #99, opened 2015-03-23
Fix: hook post-metas and headings early actions to wp_head instead of wp

eri-trabiccolo:
Could you please include also this small pull-request?
As I told you I was digging into displaying a static front page with pagination (template with a custom loop) and found that some of those actions hooked to wp, for some reason, reset the query var 'page', which is needed in those cases (https://codex.wordpress.org/Pagination#static_front_page) . I've found that the problem happened (for example, in headings) here: https://github.com/Nikeo/customizr/blob/master/inc/parts/class-content-headings.php#L71 (when hooked to wp). Basically the <pre>is_front_page()</pre> caused that. I cannot understand the reason though.
So not all those methods should be re-hooked to wp_head to avoid this, just the ones which refers to is_front_page(), tc__f('__is_home') and so on.
I think there's nothing wrong with that, right? I mean, no side effects..
I also hope this can fix the issue with some plugins showing custom loops in static home front page (see edd).
For sure it fixes the issue with my template ;)
But if you have any doubt  well I don't get offended :package: 

Good night ;)

### DIFF
--- a/inc/parts/class-content-headings.php
+++ b/inc/parts/class-content-headings.php
@@ -17,11 +17,11 @@ if ( ! class_exists( 'TC_headings' ) ) :
       function __construct () {
         self::$instance =& $this;
         //set actions and filters for posts and page headings
-        add_action( 'wp_head'                            , array( $this , 'tc_set_post_page_heading_hooks') );
+        add_action( 'template_redirect'                            , array( $this , 'tc_set_post_page_heading_hooks') );
         //set actions and filters for archives headings
-        add_action( 'wp_head'                            , array( $this , 'tc_set_archives_heading_hooks') );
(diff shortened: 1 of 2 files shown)


Nikeo:
OK Thanks. I guess you already tried template_redirect ?

eri-trabiccolo:
nope I'll do it now ;)
I'm on edd, in local with this change the pagination works, though on that guy installation looks like it doesnt :/

eri-trabiccolo:
template redirect works fine too in local.. I'm trying to see why doesn't work there

Nikeo:
OK! I wait for your feedback then thanks !

eri-trabiccolo:
Update:
use template_redirect hook

PR #101, opened 2015-03-25
Fix: don't print the menu wrapper when no menu set

eri-trabiccolo:
Should fix:
http://themesandco.com/support-forums/topic/adjusting-size-or-removing-empty-navbar/
http://themesandco.com/support-forums/topic/disable-3-bar-menu-on-single-page-site/

### DIFF
--- a/inc/parts/class-header-menu.php
+++ b/inc/parts/class-header-menu.php
@@ -155,33 +155,34 @@ if ( ! class_exists( 'TC_menu' ) ) :
     * @since Customizr 3.0
     */
     function tc_menu_display($resp = null) {
-      ob_start();
-        //menu setup
-        $menu_args = apply_filters( 'tc_menu_args',
-                    array(
-                      'theme_location'  => 'main',
-                      'menu_class'      => ( ! wp_is_mobile() && 'hover' == esc_attr( TC_utils::$inst->tc_opt( 'tc_menu_type' ) ) ) ? 'nav tc-hover-menu' : 'nav',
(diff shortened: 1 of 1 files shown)


Nikeo:
Hi Rocco,
What is a menu not set ? is it when no menu has been assigned to the main location of the theme ?
In this case, id you check that the fallback is still fired ?
'fallback_cb'     => array( $this , 'tc_link_to_menu_editor' )

Thanks

eri-trabiccolo:
A menu not set is .. yes when you don't set the menu to the main location of the theme.
And when the fallback is fired "there is a menu", that happens when user's logged in and can manage themes right? So in that case the wrapper is still printed.
Is not printed when you haven't set a menu to the main location of the theme, and when that fallback doesn't return anything.
I'll make another test anyway.

eri-trabiccolo:
This pr is wrong, I'm investigating on the causes :D

eri-trabiccolo:
Ah the problem is caused by the old tc_navbar_display do we want to finally get rid of it and make tc_new_menu_view be the real tc_navbar_display?

Nikeo:
ah ok. 
getting rid of the old navbar_display : I don't have a clear idea of the possible consequences and don't have the time to investigate now. Let's postpone this PR then.

eri-trabiccolo:
Ah no, was me doing a wrong test.
I update this pr

eri-trabiccolo:
Done,
but yes feel free to postpone it ;)

eri-trabiccolo:
Close this messy pr. :D

PR #102, opened 2015-03-25
Fix: avoid plugin's conflicts with the centering slides feature:

eri-trabiccolo:
replace the #customizr-slider's 'slide' class with 'customizr-slide'

Alternatively we can parametrize it (tcparams)
or yell at some developers :P

### DIFF
--- a/inc/assets/js/parts/bootstrap.js
+++ b/inc/assets/js/parts/bootstrap.js
@@ -1870,7 +1870,7 @@ var TCParams = TCParams || {};
     }
 
   , slide: function (type, next) {
-      if(!$.support.transition && this.$element.hasClass('slide')) {
+      if(!$.support.transition && this.$element.hasClass('customizr-slide')) {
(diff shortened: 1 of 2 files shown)


Nikeo:
Hi Rocco!
Thanks for this. Since the slider is a pretty sensitive thing I really want to be extremely cautious on this one, so :
1) Did you make sure that there's no side effect with this class changes, server side + browser side ?
2) Can you confirm that this has been tested with a previously cleaned browser cache ?

Thanks !

eri-trabiccolo:
Hi Nico, well 
2) should confirm 1) right? Unless I didn't miss anything.
Yes I tested it, mainly because I wasn't sure that just those 2 changes will be enough. :)
I'll test it again.

Nikeo:
OK thanks. I've made some tests too and everything looks good.
Merged

eri-trabiccolo:
:+1:

PR #103, opened 2015-03-25
Fix: fix logo ratio, apply only when no sticky-logo set

eri-trabiccolo:
Hi Nico,
pretty important 'cause the sticky-logo ratio was totally wrong ( I know it was one of my ideas :D).
With this fix we set the width/height just to the main logo, and just when no sticky-logo is set ('cause when there's a sticky-logo we don't need transitions).

Cheers

### DIFF
--- a/inc/assets/js/parts/main.js
+++ b/inc/assets/js/parts/main.js
@@ -459,7 +459,7 @@ jQuery(function ($) {
           isCustomizing   = $('body').hasClass('is-customizing'),
           customOffset    = +_p.stickyCustomOffset,
           $sticky_logo    = $('img.sticky', '.site-logo'),
-          logo            = 0 == $sticky_logo.length ? { _logo: $('img:not(".sticky")', '.site-logo') , _ratio: '' }: false;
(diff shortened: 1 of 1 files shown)


Nikeo:
Hi Rocco,
Thanks for this fix.
Can you please explain a bit further the problem solved here because : I don't get was is the sticky-logo ? Are you talking about sticky header ?
Can you confirm that you checked this code ?

eri-trabiccolo:
Hi Nico.
Nope for sticky-logo, I mean the second logo for the sticky header. Do you remember that option?
That one. So since we apply an inline style to the logo to make the transition working for the main logo when we pass from the normal header to the one when scrolling, when there are 2 different logos we will not see any "width/heigh" transition anyway. Do you agree?
So setting it is useless in that condition. Do you agree?
Plus, calculating width/height for the sticky-logo ratio results in the attr width/height, and then height set as current width / ratio, or something of the sort. The thing is that the current width for the sticky-logo (which is hidden at the start) it's not correct 'cause it's not "forced" by the css/container, so if you have a big image you'll see a very bad sticky-logo.
Anyway I tested it but can make further tests.

Nikeo:
ok perfect I agree.

PR #104, opened 2015-03-27
Fix: fix bug when force default layout was checked

eri-trabiccolo:
Important.


Nikeo:
fixed in dev. Can't believe this bug shows up only now ! :-O
Excellent catch thanks :)

eri-trabiccolo:
:)
So it's old.. ahah wow.
Well thanks to Mats and Maksim :D

PR #107, opened 2015-03-30
Fix: gallery - handle the case link to none

eri-trabiccolo:


Nikeo:
Hi Rocco, does it fix a reported problem in the forum(s) ?
What is the typical workflow to reproduce this ?
Thanks :)

eri-trabiccolo:
It fixes this issue:
http://presscustomizr.com/snippet/allow-thumbnails-links-ie10/comment-page-1/#comment-316315
(well in PC these anchors don't work) anyway the comment is one of the latest, Luca's one.
Basically when you decide to don't link the gallery image, that current method still forces the link to the media file. The fix above prevents it.
So to reproduce it, make a gallery, don't link the images, publish, click on an image ..and see.. well you shouldn't be able to click on it at all since they're supposed to don't have a link :D

Nikeo:
Ok perfect thanks. merged.

PR #108, opened 2015-03-31
Fix: fix collapsed customizer panel

eri-trabiccolo:
Thanks to Sean James Broderick for reporting this.


Nikeo:
Excellent thanks. Funny I just noticed this today as well.

eri-trabiccolo:
:)

PR #109, opened 2015-03-31
Fix: don't display the menus when no menu is set

eri-trabiccolo:


Nikeo:
Hi Rocco, could you quickly recall me what problem does this PR solve ?
Thanks (sorry you've been pushing this one a while ago :) )

eri-trabiccolo:
This was born a while ago to solve a minor issue, which is the printing of the resp menu button even if no menu is set.
Here an example of filed bug:
http://presscustomizr.com/support-forums/topic/disable-3-bar-menu-on-single-page-site/

Nikeo:
ah ok. Thanks.

eri-trabiccolo:
Yeah not very important stuff :)

Nikeo:
yep :D. but thanks for the quick reply anyway

eri-trabiccolo:
you're welcome, until the memory will allow me to remember medium-recent events :D

eri-trabiccolo:
Included in the menu revamp

PR #118, opened 2015-04-09
Fix: override bootstrap thumbnails left margin for woocommerce ones

eri-trabiccolo:
You can see the thumbnail negative margin issue.
http://demo.presscustomizr.com/product/happy-ninja/

This should fix it


Nikeo:
Ok thanks :)

eri-trabiccolo:
Ahah, this sounds mostly like "Yeah you told me about this.. but I'm not convinced" :P

PR #120, opened 2015-04-10
Fix: don't load unnecessary classes when in wordpress Customize

eri-trabiccolo:
Front classes don't need to be loaded when in customize left panel (is_admin() === true)
Same thing for pro activation classes.

I think is fine, but double check, your competence in the customize land is far better than mine, though I'm better with women .. I know :P


Nikeo:
changed in dev. Same philosophy but different implementation.

PR #121, opened 2015-04-12
Fix: fix issue with contents containing many many links

eri-trabiccolo:
Ok this is kinda hack.
Reported here: http://presscustomizr.com/support-forums/topic/maximum-number-of-hyperlinks-in-a-page/#post-68165

Looks like that with a long list of links in the content that preg_replace returns an empty string, dunno why, haven't investigated on it. Maybe we can use a better pattern/replacement? Dunno, is about some memory leak? Dunno .. but probably the latter.

Anyway this just prevents the returning of an empty string.


Nikeo:
Cool, good hack :)

PR #124, opened 2015-04-14
Fix: when not  is set disable comments

eri-trabiccolo:
Sorry for the last comment, didn't read carefully the list entry.
tested with 404 .. should be overall fine.. right?

(ops, missed "post" in the title, that's 'cause I used the dollar .. :/)


Nikeo:
good thanks !
merged in dev

PR #125, opened 2015-04-14
Fix: limit the expansion effect on article.hover just to tc-thumbnail el...

eri-trabiccolo:
...ements


Nikeo:
Does it fix a particular problem reported somewhere ?
Thx

eri-trabiccolo:
Not really reported but.. for example if you want to use a custom hook for fpu inside an article ... well that rule will expand the fps as soon as you enter in the whole article (there's an fpu related pull-request in which I explain it)

Nikeo:
OK, I agree.
Merged in dev:)

PR #127, opened 2015-04-16
Add: tc-mainwrappers methods for plugin compatibilities

eri-trabiccolo:
Matter of extensibility .. I think is good 'cause we might use those for other plugins.
Maybe those methods can be moved in utils, dunno ;)

### DIFF
--- a/inc/class-fire-plugins_compat.php
+++ b/inc/class-fire-plugins_compat.php
@@ -281,46 +281,12 @@ if ( ! class_exists( 'TC_plugins_compat' ) ) :
 
       function tc_woocommerce_wrappers() {
         switch ( current_filter() ) {
-          case 'woocommerce_before_main_content':
-
-          ?>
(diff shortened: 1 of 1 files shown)


Nikeo:
Hi,
just took a look 
1) why aren't you referecing the instance with $this ?
2) did you test this new code with the latest version of WooCommerce ?

Thanks
For future references and since I often merge you PR with a delay :) , it would be a good practice to always specify if the PR has been tested (with the relevant plugin if impact plugin).

eri-trabiccolo:
Right ;)
1) 'cause inside that function $this is unreachable (out of scope, change that to $this and you'll see the error)
2) Well now I'm testing it (and it's fine) but the thing is, if this doesn't work neither the first code should :D 
Anyway I see there's a problem with the comments in the shop page when you have no products (the same you have in 404pages, well you already merged that in dev so .. fine :D)

Nikeo:
ok.
Comment => this fixes 2 problems (at least) then. great!

PR #129, opened 2015-04-16
Fix: don't print private taxonomies in post metas tags

eri-trabiccolo:
Well the better thing would be to don't retrieve them at all if that filter returns false, but get_oject_taxonomies doesn't allow that... right?


eri-trabiccolo:
We have at least 4 people complaining for this (it happens with polylang private tax)

Nikeo:
Thanks.
Merged in dev

eri-trabiccolo:
Ok, this is tricky so .. I have to better check this when you'll merge it :P

PR #130, opened 2015-04-17
Fix: cache and use cached common jquery elements

eri-trabiccolo:
What do you think about this?
Should be better in terms of performances.
Better if we globally cache common elements like window, body etc directly in .. say tc-js-params.js
... ?
Well not there 'cause it's outside jquery, but maybe in main.js and using just one enclosure .. maybe


eri-trabiccolo:
The second commit 'cause as for now we consider just the border-bottom.

Nikeo:
fully agree
good move
merged in dev

PR #131, opened 2015-04-17
Fix: disabling global tc_post_metas didn't hide metas

eri-trabiccolo:
When you disabled "Display post metas" in Customize, metas were hidden in the preview but this had no real effect in the live site.
This should fix that, also should better handle the Customize preview.
(hopefully I am not that sick .. so double/triple check this ;) )

p.s.
reported in the pro forum


Nikeo:
looks good thanks :)
Merged in dev

PR #134, opened 2015-04-21
Add: basic buddypress support (don't show comments in buddypress pages)

eri-trabiccolo:
Look that we have also some code for the compatibility of a premium plugin which works with buddypress. Do we want to add this?

We also have the optimizepress comp, but I don't know what will be with all these PR on the same file..
(will push it in some hours if you don't tell me to do otherwise :P)

might conflict with #132 and #127


Nikeo:
looks perfect.
Merged in dev.

PR #135, opened 2015-04-21
Gallery options

eri-trabiccolo:
Added 3 new options
Removed now useless code rewriting of the original gallery rendering code.

### DIFF
--- a/inc/admin/js/parts/_control.js
+++ b/inc/admin/js/parts/_control.js
@@ -254,6 +254,15 @@
       callback: function (to) {
         return '1' == to;
       }
+    },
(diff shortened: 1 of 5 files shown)


eri-trabiccolo:
p.s.
wp since 3.9 adds optional html5 gallery and captions support.
we might need to write some CSS to handle them and eventually add theme support to them

Nikeo:
Good addition. A gallery section was missing.
Clean. thanks. Merged in dev.

eri-trabiccolo:
As side good effect will make the transition work for galleries in the customize ;) (stronger selector will override some rules which prevented that;))

PR #136, opened 2015-04-21
Fix: disable fade hover links for first level menu items in ie

eri-trabiccolo:
I was able to reproduce the issue with ie11 and the demo site, on which I modeled the fix (with inspect element live) but wasn't able to reproduce it on my local installation, so this requires a double check.
The fix is extended on all ie versions, really couldn't test them all :D 
Still confused about why this happens just with those specific links.. mah
https://wordpress.org/support/topic/navigation-hover-does-not-work-in-ie11?replies=1
( though the css there was buggy :D)


Nikeo:
cool. merged in dev

PR #138, opened 2015-04-22
Fix: display unknown archive types headings;

eri-trabiccolo:
use if/else statement when retrieving archive headings/classes
immediatily return the archive class when asked for and achive matched

for something reported here:
http://wordpress.org/support/theme/customizr

### DIFF
--- a/inc/parts/class-content-headings.php
+++ b/inc/parts/class-content-headings.php
@@ -178,8 +178,8 @@ if ( ! class_exists( 'TC_headings' ) ) :
           return $_title;
 
         //gets the post/page title
-        if ( is_singular() || ! apply_filters('tc_display_link_for_post_titles' , true ) ){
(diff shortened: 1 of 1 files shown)


Nikeo:
OK
Did you managed to check this with all archive cases ?
What was the initial problem to treat if you can remember it? (I can't find the related topic...)
thanks Rocco

eri-trabiccolo:
I checked that with all the archive type cases I could think about :D so tags authors and so on.
Well I pasted a very useful link for that topic ahah, sorry.. was this -> https://wordpress.org/support/topic/taxonomy-archive-how-to-display-title-and-description#post-6855755

That's a typical case.
Consider that I looked at the twentyfifteen archive.php to see why they worked fine :D .. So if it's an archive and isn't one of them we take care of just "fallback" on the common way to display them. That's the sense.
What do you think?

Nikeo:
ok. looks good !
merged in dev

eri-trabiccolo:
what's that dancing emoticon? well think about I used it :P

PR #139, opened 2015-04-23
Fix: remove reference to the tag, use site-description tag

eri-trabiccolo:


Nikeo:
yep optimized.
merged in dev

PR #140, opened 2015-04-23
Fix: don't use amatic weight of 700, workaround for missing question mark

eri-trabiccolo:
Sorry for the title, of the commit, while founding the solution I found that "nothing" is better than 400 (it still allows a bold size)


Nikeo:
ok fine.
merged in dev
thx !

PR #142, opened 2015-04-28
Fix: revert some references to old function names

eri-trabiccolo:
:)
Do you agree?


Nikeo:
Looks good!
tested on your side ?

eri-trabiccolo:
yeppe :D

PR #143, opened 2015-04-29
Fix: use the correct post id when retrieving the grid layout

eri-trabiccolo:
Problem:
When you have a blog page with a specified layout different than the global layout we still retrieve the global layout.
This happens 'cause we use get_the_ID() even when we're in the loop (__before_article is in the wordpress loop), even using TC_utils::tc_id() will not solve this, 'cause we return get_the_ID() if we're in the loop. So my solution is to retrieve the object id as soon as possible (set it as private attribute of the class) and use it every time we need to retrieve the layout.

### DIFF
--- a/inc/parts/class-content-post_list_grid.php
+++ b/inc/parts/class-content-post_list_grid.php
@@ -15,6 +15,7 @@ if ( ! class_exists( 'TC_post_list_grid' ) ) :
     class TC_post_list_grid {
         static $instance;
         private $expanded_sticky;
+        private $post_id;
 
         function __construct () {
           self::$instance =& $this;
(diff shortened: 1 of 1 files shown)


Nikeo:
Best solution. :)
included in a 'wp' callback so the query object is ready and well formed.
Perfect for me.
ready for merging!
thx

eri-trabiccolo:
:)
Yes, maybe we can consider, in the future to retrieve the queried object id by default in some of our classes (class-fire-init ?) so we can access everywhere we need it. Will see

Nikeo:
yes that could be useful. And use it in TC_utils::tc_id()

eri-trabiccolo:
something like (if not in the loop) if TC_init::$object_id (static) return it otherwise "build and assign it to our static property" ?

Nikeo:
yep something like that

PR #147, opened 2015-05-01
Add: comment anchor link

eri-trabiccolo:
Basic comment anchor link.
The whole comment displaying is probably a little messy :D 
I used sizes expressed in "em", according with the other "icons":before

The "real" problem is that the smooth scroll doesn't work of course, because the href doesn't start with "#".. My idea is that handle this at this stage is pretty useless, we can handle it in the new smooth scroll.
Do you agree?

p-r info:
might conflict with #146 (both affects tc_custom.less)
against v3.3.21 build v5

p.s.
bonus, comment's text-area as border-box. Width 100% + padding made it overflow the window in smaller viewports.
Probably we should make all the text-areas / input text, this way (?)

### DIFF
--- a/inc/assets/less/tc_custom.less
+++ b/inc/assets/less/tc_custom.less
@@ -1523,6 +1523,9 @@ article .author-info h2 , .archive-header .author-info h2 {
 
 #respond #comment {
   width: 100%;
+  -webkit-box-sizing: border-box;
+  -moz-box-sizing: border-box;
+  box-sizing: border-box;
 }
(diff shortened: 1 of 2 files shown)


eri-trabiccolo:
Visible just on hover?
.tc-comment-anchor {
    visibility: hidden;
}

article.comment:hover .tc-comment-anchor {
    visibility: visible;
}
also, probably if we display this, then we should remove the anchor link in the comment-date..

PR #148, opened 2015-05-03
Fix: don't print private taxonomies, with fix for old php versions

eri-trabiccolo:
Problem:
looks like in old php versions (< 5.3) we cannot access the property "public" of that object. I feel 'cause it ( "public") is somewhat "reserved", as it doesn't happen with the property "hierarchical"

Fix:.
Cast the stdClass object to an array in order to access to its key "public". 
Tested with php 5.2.8, do you have the chance to double check it.. or maybe you have some other idea.

against 3.3.22 build

### DIFF
--- a/inc/parts/class-content-post_metas.php
+++ b/inc/parts/class-content-post_metas.php
@@ -392,8 +392,13 @@ if ( ! class_exists( 'TC_post_metas' ) ) :
 
           //filter the post taxonomies
           while ( $el = current($tax_list) ) {
+              // cast $el stdClass object in an array to access its property 'public'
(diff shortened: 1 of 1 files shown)


Nikeo:
Ah this is it then. I have the same "feeling". I don't see any other possible reason anyway since the syntax of the if statement is correct and there's no formal php error in the way variables are declared / used.

A possible enhancement would be to create a helper method to replace this quite long condition in the if !
Thanks for the fix :)

eri-trabiccolo:
You know how much I love make helpers :D But in this case .. dunno we should pass to many params (even if we can use an array) dunno if it's worthwhile ..

PR #149, opened 2015-05-03
Fix: don't display tax archive titles for some plugins

eri-trabiccolo:
Against 3.3.22 build


Nikeo:
Merged in dev
Thanks :)

PR #150, opened 2015-05-04
Fix: don't show slider in home when no home slider is set

eri-trabiccolo:
https://wordpress.org/support/topic/main-page-slider-trouble?replies=2#post-6906522


Nikeo:
Thanks for this PR. :)

I had to fix another problem afterwards due to the check 0!= $tc_front_slider. It had to be replaced by !empty($tc_front_slider) to work properly.

eri-trabiccolo:
right... sorry :/

PR #153, opened 2015-05-08
Fix: properly filter get_the_content() for special post formats

eri-trabiccolo:
Bug reported here:
https://wordpress.org/support/topic/in-category-view-shortcodes-not-working

Fix against a655944

tested by myself with contact-form shortcode inside an aside post format both with alternate layout and grid (just to test it didn't affect the grid :D rock --pedantic mode)

p.s.
Why we show the pagination in post lists? (nextpage)


Nikeo:
OK fine. thanks
I will just rename the callback with a more explicit name like, fix_shortcode_characters or smtg like that

eri-trabiccolo:
Ah ok, well I used a name which was similar to our "tc_the_title" :D

eri-trabiccolo:
Ah you meant the callback title, well easily if you say "I'll just rename the callback" you really mean that :D And of course you are right ;)

PR #154, opened 2015-05-08
Fix: remove link in post thumbs

eri-trabiccolo:
Why do we wrap the image in a link to the same page?
mmm

against a655944574


eri-trabiccolo:
Why don't we add another location option?
As for now we have before the title in full width, before the title boxed (which means inside the #content) and after the title (still inside the #content). I think we miss before the title boxed should be like before the title in f-w but boxed, so in a container.
So we can say:
1) before the title in full width => after the header in full width
2) (new) => after the header boxed
3) before the title boxed => before the title
4) after the title => remains, after the title

What do you think?

PR #156, opened 2015-05-09
Fix: remove smartload noscript tag

eri-trabiccolo:
For a bug reported here: http://presscustomizr.com/support-forums/topic/phantom-images-in-manual-slideshow/

Temporary fix or just a reminder..

against 92eda6b681


Nikeo:
Yes good. This has to be removed now. I agree :)

eri-trabiccolo:
:+1:

PR #157, opened 2015-05-10
FIX : recursive creation of iphonecheckcontainer

JTS-IL:
On post/page metas, the slider for page selection , after clicking the YES/NO iphonechacheckbox it generates a new div "iphonecheckcontainer" nested under "iphonecheckcontainer". it reproduces after every checked change, and happens only here.
This "funny" change - wrapping the inputs with div, does the fix.

Maybe there's a better way to fix it in the js functions?


JTS-IL:
Reported by me here: https://wordpress.org/support/topic/multiple-instance-of-iphonecheck

JTS-IL:
tested again - something went wrong. closing PR

PR #158, opened 2015-05-11
Fix: remove btt-arrow inline style, rule moved in the skin css

eri-trabiccolo:
Well, no great comments to do :)

against ec2da9e890


Nikeo:
Yes good PR. This will allow the customizer live preview when hovering the skins.
Thanks :)

eri-trabiccolo:
Yep,
I noticed it while testing the random skin but then I forgot it :D

PR #159, opened 2015-05-14
Fix: whene deleting retina images don't forget the original attachmen…

eri-trabiccolo:
…t's retina version

Reported here:
http://presscustomizr.com/support-forums/topic/2x-version-image-is-not-deleted/
Basically the retina version of the original attachment (base size) isn't removed, that's cause we cycle on the sizes metadata of the attachment.

What this PR does:
1) remove the check on retina option enabled, 'cause I think we have to remove them even if that option is disabled. Scenario:
- Retina enabled ; media uploaded; retina files created
- user disable the retina option; then delete an attachment
- retina images aren't deleted. 
I think is wrong

2) Add the original attachment file to the list of the "sizes" to cycle through

3) Simplify the cycle: cycle only on the sizes metadata.

against 54c965cd59

### DIFF
--- a/inc/class-fire-init.php
+++ b/inc/class-fire-init.php
@@ -634,28 +634,24 @@ if ( ! class_exists( 'TC_init' ) ) :
      * @credits http://wp.tutsplus.com/author/chrisbavota/
      */
       function tc_clean_retina_images( $attachment_id ) {
-        //checks if retina is enabled in options
-        if ( 0 == TC_utils::$inst->tc_opt( 'tc_retina_support' ) )
-          return;
-
         $meta = wp_get_attachment_metadata( $attachment_id );
         if ( !isset( $meta['file']) )
           return;
 
         $upload_dir = wp_upload_dir();
         $path = pathinfo( $meta['file'] );
-        foreach ( $meta as $key => $value ) {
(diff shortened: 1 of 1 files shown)


Nikeo:
Looks good.
Thanks for those very clear explanations.
Quick question : did you test different cases ?

eri-trabiccolo:
Tested removal of an image. Cannot think at different cases, for example?
The only real difference from the original code is:
1) the check on retina enabled removed. In any case there's a pre-existent  final check on the file existence.
2) Add the orginal file name as a "size".

Loop changes shouldn't be subject to "different cases" :) Or maybe we can have a meta without sizes? So maybe we can check if 'sizes' key exists, so replace : 
$sizes = $meta['sizes']; 
with
$sizes = array_key_exists('sizes', $meta) ? $meta['sizes'] : array();

right?

Nikeo:
OK so that"s right I don't think there are "cases" so to speak either.
Just wanted to know if you made a test on your side already :)

eri-trabiccolo:
Yep I tried removing a media :D 
Don't think we need that check on the 'sizes' key existence, it might be empty but still should be present :)

PR #160, opened 2015-05-15
Sticky header trigger height

eri-trabiccolo:
This is my idea to avoid the huge gap when starting the scroll below the sticky header.
I set the tiggerHeight as half the marginTop, maybe we can change it, in any case I think it should be, if not custom, a percentage of the #tc-reset-margin-top's margin-top . (maybe 2/3 could be even better when logo centered).

You might like or not this but there are two (independent) additional fixes in it (I might do another PR but since are 2 lines..).

Scenario for the bootstrap.js change:
- adminbar on , tc-sticky-header on, dropdowntoViewport off, so "In responsive mode, limit the height of the dropdown menu block to the visible viewport" unchecked. 
- scroll down the page for a certain amount (so the header top will be set to 0), click on the menu button, page will scrolled up and menu will be shown (sticky disabled)
- click again on the menu, sticky re-enabled but top is set to 0 so the adminbar will overlap the header, or if you prefer the header will go behind the adminbar :D 
I trigger a resize so the header's offset will be set again. We might consider to make custom signals so we can better control this stuff without trigger a window resize.

Scenario for "initialOffset = ! $_window.scrollTop() ? $wpadminbar.height() : 0;" change:
- adminbar on, tc-sticky-header on
- scroll down the page just ONE step, you'll see the sticky header enabled having an useless top property = adminbar height, so blank gap

My solution is to refer to the window scrollTop in that case instead of relying on _is_scrolling()

### DIFF
--- a/inc/assets/js/parts/bootstrap.js
+++ b/inc/assets/js/parts/bootstrap.js
@@ -1687,6 +1687,7 @@ var TCParams = TCParams || {};
       //@tc adddon
       if ( TCParams && 1 != TCParams.dropdowntoViewport && 1 == TCParams.stickyHeader ) {
         $('body').addClass('tc-sticky-header');
+        $(window).trigger('resize');
       }
     }
 
--- a/inc/assets/js/parts/main.js
+++ b/inc/assets/js/parts/main.js
@@ -495,9 +495,9 @@ jQuery(function ($) {
         $resetMarginTop.css('margin-top' , '' ).show();
 
         //What is the initial offset of the header ?
-        var headerHeight    = $tcHeader.outerHeight(true); /* include borders and eventual margins (true param)*/
+        var headerHeight    = $tcHeader.outerHeight(true), /* include borders and eventual margins (true param)*/
             //initial margin-top = initial offset + header's height
-            marginTop       = initial headerHeight + customOffset;
+            marginTop       = headerHeight + customOffset;
         //set initial margin-top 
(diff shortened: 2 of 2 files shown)


inc/assets/js/parts/bootstrap.js
```diff
@@ -1687,6 +1687,7 @@ var TCParams = TCParams || {};
       //@tc adddon
       if ( TCParams && 1 != TCParams.dropdowntoViewport && 1 == TCParams.stickyHeader ) {
         $('body').addClass('tc-sticky-header');
+        $(window).trigger('resize');
```
eri-trabiccolo: First scenario

inc/assets/js/parts/main.js
```diff
@@ -479,7 +480,7 @@ jQuery(function ($) {
             if ( 580 < $_window.width() )
                 initialOffset = $wpadminbar.height();
             else
-                initialOffset = ! _is_scrolling() ? $wpadminbar.height() : 0;
+                initialOffset = ! $_window.scrollTop() ? $wpadminbar.height() : 0;
```
eri-trabiccolo: Second scenario

eri-trabiccolo:
Why do I proposed this? :D Pretty *hitty since the header is fixed :D 

I will close this and propose just the bug fixes tomorrow.

About the sticky header overall we need something different. Like set it static when not scrolling and fixed when scrolling (beyond the trigger height). Probably we don't lose the transitions which are basically on the height and background.. We'll see

PR #161, opened 2015-05-15
Fix: use proper priority for tc_parse_imgs callback of the_content fi…

eri-trabiccolo:
Re enable lazy load in galleries.

Tested in normal posts, gallery post formats, normal posts with galleries.

Work fine and shouldn't revert anything.

Dunno why this has stopped to work though..

### DIFF
--- a/inc/class-fire-utils.php
+++ b/inc/class-fire-utils.php
@@ -60,7 +60,6 @@ if ( ! class_exists( 'TC_utils' ) ) :
         if ( esc_attr( TC_utils::$inst->tc_opt( 'tc_img_smart_load' ) ) ) {
           add_filter( 'the_content'                       , array( $this , 'tc_parse_imgs' ), 20 );
           add_filter( 'tc_thumb_html'                     , array( $this , 'tc_parse_imgs' ) );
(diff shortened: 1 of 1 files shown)


eri-trabiccolo:
Ah I see, 'cause before the gallery rewrite we do use the post_gallery filter so it worked.
Anyway since galleries are still part of the content we can(should) get rid of this:
https://github.com/Nikeo/customizr/blob/dev/inc/class-fire-utils.php#L63
Because plugins might use the gallery shortcode somewhere else and the smartload js will still work just on '.article-container, .__before_main_wrapper, .widget-front'
(think for example to a plugin which uses the gallery shortcode + that filter in widgets not inside the above elements).

So finally "Dunno why this has stopped to work though.." is "I know, it's my fault, sorry ;)"

Nikeo:
cool. no clue either

eri-trabiccolo:
My fault, since I got rid of the old gallery code which was hooked on the post_gallery filter hook the smart load didn't work anymore with them. 'Cause code in wp-includes/media.php doesn't pass the actual gallery content to it but ''. It is something you can hook to not override the normal gallery but to use your gallery preventing the wp gallery being elaborated and printed.
So it was my fault.

PR #164, opened 2015-05-17
Fix: check if get_term_link returns a WP_Error: compatibility with ol…

eri-trabiccolo:
…d wp

against 572e756

Following to this https://wordpress.org/support/topic/empty-articles-when-upgrading-to-customizr-version-332

I found that at least wp 3.6.1 
[get_term_link($term->term_id, $term->taxonomy)]( https://github.com/Nikeo/customizr/blob/dev/inc/parts/class-content-post_metas.php#L366 )

returns a WP_Error 
Looking at the [codex](https://codex.wordpress.org/Function_Reference/get_term_link#Examples), looks like we can just use get_term_link($term), when $term is a term object. Just this change avoids the issue with 3.6.1, but I thought should be better make a check anyway on the return type of that function.

I tested it and looks we don't have issues... but, don't know, maybe I'm missing something.


Nikeo:
Thanks for catching this.
Looks fine for merging in dev.

eri-trabiccolo:
up

PR #168, opened 2015-05-21
Fix: exclude page and attachment post types when including cpt in arc…

eri-trabiccolo:
Following to:
https://wordpress.org/support/topic/custom-post-type-in-search-and-author-archive-and-date-archive?replies=21

The problem is that when we include cpt in archives we set the query var 'post_type' as the array retrieved from get_post_types.
That array contains also page and attachment post types.

With this code:
1) we get just "custom post types" instructing get_post_types to exclude buit-in types
2) we add to this retrieved array the standard wordpress post type.

### DIFF
--- a/inc/parts/class-content-post_list.php
+++ b/inc/parts/class-content-post_list.php
@@ -345,7 +345,10 @@ class TC_post_list {
       return;
 
     //filter the post types to include, they must be public and not excluded from search
-    $post_types     = get_post_types( array( 'public' => true, 'exclude_from_search' => false) );
+    //we also exclude the built-in types, to exclude pages and attachments, we'll add standard posts later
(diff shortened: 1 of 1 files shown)


Nikeo:
Looks fine. thanks

eri-trabiccolo:
About this there's a minor issue related. Basically wordpress groups posts of the same type when they have the same publishing date.
Reported here:
https://wordpress.org/support/topic/getting-general-posts-in-custom-post-archive

I don't think it's a real issue, it's how wp works, I went with a small SQL add, but I'm not sure it's very fine, it probably should be:

`function _my_orderby( $orderby ) {
    global $wpdb;
    $_my_orderby = "{$wpdb->posts}.ID DESC";
    return $orderby ? $orderby . ',' . $_my_orderby : $orderby ;
}`

'cause when isn't set an original orderby we shouldn't force it .. I think

PR #175, opened 2015-05-30
Fix: better rtl slide controls and swiping(js)

eri-trabiccolo:
Following to this:
https://wordpress.org/support/topic/slider-problem-after-last-update

1) slider class:
old code basically didn't change the slid direction, as you can see "left" meant always "prev" and "right" always "next". In rtl right -> prev , left -> next , so we have to change just the left/right class.

2)js swipeleft/right: (still on the old main.js)
if I get the issue, swiping to the left, in rtl, should replicate "prev" and swiping to the right "next"

@JTS-IL consultancy would be very helpful :)

### DIFF
--- a/inc/assets/js/parts/main.js
+++ b/inc/assets/js/parts/main.js
@@ -127,11 +127,12 @@ jQuery(function ($) {
         });
 
         $('.carousel' ).each( function() {
+            var _is_rtl = $('body').hasClass('rtl');
             $(this).hammer().on('swipeleft tap', function() {
-                $(this).carousel('next');
+                $(this).carousel( ! _is_rtl ? 'next' : 'prev' );
             });
             $(this).hammer().on('swiperight', function(){
-                $(this).carousel('prev');
+                $(this).carousel( ! _is_rtl ? 'prev' : 'next' );
             });
         });
     }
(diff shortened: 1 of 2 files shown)


JTS-IL:
I'm not sure I understand the fix in js.
The correct direction is to slide from left to right, as paging in a book

JTS-IL:
On second thought - I think that it's OK as it is on 3.3.26
Exactly as on LTR the left arrow slides left and the right slides right, the functionality of the arrows does not change since the direction should be absolute.

what do you think?
@Nikeo ?

eri-trabiccolo:
I don't think it's OK as it is on 3.3.26, but well it's just my opinion of ltr addicted :D

The sense of this PR is this:
In RTL
desktop:
1) when you click on the right control ">" you go in the "past", so you show the prev slide
2) when you click on the left control "<" you go in the "future", so you show the next slide
mobiles:
1) swiping right you show the next slide, in LTR you swipe left to show the next slide
2) swiping left you show the prev slide, in LTR you swipe right to show the prev slide

this is consistent with the animation directions and with what that user wanted (I think).

JTS-IL:
OK.
Now I got it! (was quite stupid not to see it before)
I added numbers to my slider and understood what you mean :-)

eri-trabiccolo:
> I added numbers to my slider and understood what you mean :-)
> - JTS-IL

I did the same :D

Nikeo:
merged in dev.
js part merged in _main_slider.part.js

eri-trabiccolo:
mm cannot see the merge, don't forget it :P

PR #176, opened 2015-06-01
Add: new option - sticky footer

eri-trabiccolo:
Hello,
this is a beta, I need to make further tests on ie + safari asap.
I made the PR anyway in order to let you see how it was taking shape.
Talk you soon.

p.s.
against new-default-menu branch because of the #tc-page-wrap


Nikeo:
OK Thanks.
This will be merged after the new JS framework and the new menu

eri-trabiccolo:
will be re-done according to the new JS framework

PR #179, opened 2015-06-14
Fix: use original img sizes

eri-trabiccolo:
against 4994c9

What this PR affects to:
1) Sticky logo transition:
1.a) re-introduce the logo ratio computation in the sticky header part (was lost due to a typo in the recent front revamp) to allow logo transition, just when the browser supports css transitions
1.b) don't set the height on page load, will be set later on 'resize' event fired by the sticky header
2) Retrieving of the natural/original img sizes:
Both in the logo ratio maths above and when centering hidden slider's slides we refer to the img attributes width and height. The problem is that jetpack's photon (CDN) doesn't provide these attributes. Also retina.js messes with them. We introduce then a new jquery "function" to retrieve the natural sizes, cross-browser. My first thought was to always fallback on these attributes when present whether or not the naturalWidth/Height was available. The problem is that retina.js serves images with sizes in pixels different from the attribute values. Basically, for example, attr width != naturalWidth . Since this happens just with retina imgs I left the fallback on attributes just when the browser's doesn't support naturalWidth/Height (old browsers like ie8 and below)  which pretty unlikely will run on retina devices ... I hope :D

### DIFF
--- a/grunt-tasks-config/concat.js
+++ b/grunt-tasks-config/concat.js
@@ -7,7 +7,7 @@ module.exports = {
     dest: '<%= paths.front_js %>parts/main.js',
   },
   front_js: {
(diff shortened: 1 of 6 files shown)


Nikeo:
excellent PR Rocco. Thanks :D

eri-trabiccolo:
Thanks, but the dynamic plugin creation is not my merit :) I just modded very slightly this one:

http://www.jacklmoore.com/notes/naturalwidth-and-naturalheight-in-ie/
I refer to it in the plugin description :)

PR #180, opened 2015-06-14
Fix: prevent hiding of selecter dropdown

eri-trabiccolo:
reported here: http://presscustomizr.com/support-forums/topic/slider-not-available-in-select-slider-drop-down

What do you think? Can be enough? Is this something you might like? Do you think we can have browser's issues?

Take your time to reply to all these questions :P


Nikeo:
I actually had this in mind too.
Don't see any potential bad side effect for now. So let's do it.
Thanks

PR #186, opened 2015-06-19
Emoji

eri-trabiccolo:
Should work fine.
Needs your check though.

With this code  we also load the holder.js only when needed.


Nikeo:
you've already moved modernizr in the head !! :)

eri-trabiccolo:
Sorry, this is a typo ... too many branches here :D 
You better move it to the footer again... :/

Nikeo:
ah
mmm is it the only one? or could there be others :/

eri-trabiccolo:
The only one typo coming from another branch yep.
But about the classic plugin we relies on that plugin's code, I've tested that the old emoj work fine and that the new emoj are used instead, when you don't want to display fp images and so on.
I've changed the plugins code in order to not affect "admin" part also.
That's basically 'cause I use some TC_featured_pages helpers which as you know are unavailable when is_admin. ...

Nikeo:
ok looks perfect. thanks :)

eri-trabiccolo:
:)

PR #188, opened 2015-06-21
Update tc_custom.less

JTS-IL:
Few RTL fixes since 3.3.26


Nikeo:
merged in dev.
Thanks :)

PR #191, opened 2015-06-22
Fix: cache slider model

eri-trabiccolo:
Reason:
We call tc_get_slider_model() more than once, so why don't cache the model the first time?
New filter to easily manipulate the slides array

If you agree with this can you merge it asap? I might make some other changes soon.

(I will push a similar caching for the featured pages, now that we retrieve the thumbs 2 times [see the check on whether or not enqueue holder.js], once I've done with my devs)

### DIFF
--- a/inc/parts/class-content-slider.php
+++ b/inc/parts/class-content-slider.php
@@ -15,7 +15,7 @@ if ( ! class_exists( 'TC_slider' ) ) :
 class TC_slider {
 
   static $instance;
-  private $slider_model;
+  private $sliders_model;
 
   function __construct () {
     self::$instance =& $this;
@@ -186,9 +186,6 @@ class TC_slider {
   *
   */
   private function tc_get_slider_model() {
-    if ( isset( $this -> slider_model ) )
-      return $this -> slider_model;
-    
     global $wp_query;
 
     //Do we have a slider to display in this context ?
@@ -203,6 +200,9 @@ class TC_slider {
     if ( ! $this -> tc_is_slider_active( $queried_id) )
       return array();
 
+    if ( ! empty( $this -> sliders_model ) && is_array( $this -> sliders_model ) && array_key_exists( $slider_name_id, $this -> sliders_model ) )
+      return $this -> sliders_model[ $slider_name_id ];
+
     //gets slider options if any
     $layout_value                 = tc__f('__is_home') ? TC_utils::$inst->tc_opt( 'tc_slider_width' ) : esc_attr(get_post_meta( $queried_id, $key = 'slider_layout_key' , $single = true ));
     $layout_value                 = apply_filters( 'tc_slider_layout', $layout_value, $queried_id );
@@ -214,9 +214,9 @@ class TC_slider {
     //get slides
     $slides                       = $this -> tc_get_the_slides( $slider_name_id , $img_size );
 
-    $this -> slider_model         = compact( "slider_name_id", "slides", "layout_class" , "img_size" );
+    $this -> sliders_model[ $slider_name_id ]         = compact( "slider_name_id", "slides", "layout_class" , "img_size" );
 
-    return $this -> slider_model;
+    return $this -> sliders_model[ $slider_name_id ];
   }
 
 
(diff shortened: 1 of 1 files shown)


Nikeo:
What if there are several sliders requested in the same page ?
Did you test this case ? (rare I agree... :) )

Nikeo:
A cleaner way would be to create an class instance by slider, don't you think ?

eri-trabiccolo:
I didn't test this 'cause ... well how could there be different sliders requests at the moment?
Also we really don't handle multiple sliders, see unique id).
It should be possible if someone filters tc_slider_name_id just the second time, checking on wp_current_filter? More than rare then, but yes.
I just thought to avoid some additional "job" for the current code. To allow multiple sliders I think the code must be modded in several parts..
But maybe I'm wrong.

Yes I agree we could store models per slides but since we get the slider id in that method itself, we should do it here:
https://github.com/Nikeo/customizr/blob/master/inc/parts/class-content-slider.php#L198

Ok stash this then, we'll think better about it when we'll make the code really ready for multiple sliders.. I think :D

Nikeo:
I know it's rare but there might be users already having multiples sliders in a page and they might be in trouble.
The overall idea of caching the slider model is good and I think this has to be implemented in V4 :)

eri-trabiccolo:
Yes I agree, but I don't think there are these users :D 
So do you want a version whit an array of models in ... how many minutes? :D

Nikeo:
no don't worry about this for the moment. It can wait

eri-trabiccolo:
Last pr should handle it fine, what do you think?
We can keep this here for future references.

Nikeo:
Looks perfect !!
Did you test it a little bit ?

eri-trabiccolo:
Yep, that's my test:
`add_action( 'wp',
  function(){  
    add_filter('tc_slider_name_id', function($slider){ global $wp_current_filter; if ( in_array('__before_footer', $wp_current_filter ) )  return 'demo'; return $slider; });
    add_action('__before_footer', array( TC_slider::$instance, 'tc_slider_display' ) );
  }
);`

Two sliders printed fine, though js doesn't work at all :D 
But this is another problem (it didn't work anyway) :D

eri-trabiccolo:
Any thoughts on this?
I could put it in the main slider dev, as I said above I tested it ;)

p.s.
the reason why the carousel doesn't work correctly on the "second" slide is because of:
1) https://github.com/Nikeo/customizr/blob/master/inc/assets/js/parts/_main_slider.part.js#L31
same id for both sliders, the first one cycles the second one doesn't, not initialized
(we could use our class ".carousel-slide" which will match both without unique id issues) ... but still (read point 2)
2) controls:
https://github.com/Nikeo/customizr/blob/master/inc/parts/class-content-slider.php#L444
they refer to #customizr-slider too, so when you click on the second set of controls always the first slide slides :D
So, if you're interested, my idea is to have a static class attribute which stores the number of the "rendered" slides, and us it as the "custoimzr-slider" id suffix. We could also use the slider-name instead, but what if for some reason one wants to display the same slider twice? Edge case, of course, but still.

So let me know ;)

Nikeo:
I'm ok with your solution. Generating unique slider id's that we use as id attributes.
The sliders could then be easily be fired in the js code by looping on the various id or on the id containing "customizr-slider" string  ( something like $('div[id*="customizr-slider"]')

It also solves the control issue.

About the edge case, same slider twice, let's put it out of the scope for now ? Or the solution would be to always check the unicity of the slider's id when adding it. Possible but requires a little bit more coding with a unicity check helper.

eri-trabiccolo:
Ok, with incremental ids we should solve the edge cases too 'cause I was thinking to generate(increment) it just when displaying it, so more as field of the "view" model than the slide model.
Thanks for the quick reply.

Nikeo:
yep sounds fine :)

eri-trabiccolo:
Look, I'll use a static attribute for the $sliders_model (models caching) too, I know that we basically have all singletons at the moment, but for a matter of correctness + future changes, I think it's better make it so from the start.

eri-trabiccolo:
Will be included in the next slider pr

PR #192, opened 2015-06-23
Fix: fix external link icon on multiple occurrences and exclude parents

eri-trabiccolo:
While finding a solution to the bug reported [here](http://presscustomizr.com/support-forums/topic/single-post-page-header-is-not-transparent-when-scroll-on-others-pages-its-ok)
I've decided to add a small code to exclude also links which have a parent matching (id or class) the skip selectors param (_is_selector_allowed -> //has a forbidden parent?).

Let me know if you don't agree with it and I'll provide another PR without it ;)

The issue reported occurred because in the mailto subject there were quotes and double quotes. And those hurt:
```$( 'a[href*="' + this._href +'"] + .tc-external' )```

to avoid string manipulation in order to escape them I decided to just use the [$.next()](https://api.jquery.com/next/) on the already retrieved element (our link)

For a matter of pure [serendipity](https://en.wikipedia.org/wiki/Serendipity) the latest fixes also the bug which occurred when we had many occurrences of the same link, because in that case this check
 
``` if ( this.options.addIcon && ! $( 'a[href*="' + this._href +'"] + .tc-external' ).length )```

failed, 'cause the href was the same, so no external icon was added.

Bests

### DIFF
--- a/inc/assets/js/parts/jqueryextLinks.js
+++ b/inc/assets/js/parts/jqueryextLinks.js
@@ -22,19 +22,22 @@
 
 
     Plugin.prototype.init = function() {
+      var $_external_icon = this.$_el.next('.tc-external');
       //if not eligible, then remove any remaining icon element and return
       //important => the element to remove is right after the current link element ( => use of '+' CSS operator )
       if ( ! this._is_eligible() ) {
-        if ( $( 'a[href*="' + this._href +'"] + .tc-external' ).length )
-          $( 'a[href*="' + this._href +'"] + .tc-external' ).remove();
+        if ( $_external_icon.length )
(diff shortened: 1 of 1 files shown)


Nikeo:
Good, serendipity...

Nikeo:
clean PR thanks

eri-trabiccolo:
Ops, I forgot something here.
Do we want to add something like a compatibility for that user's plugin?
I mean most likely the sharing links will have the external icon, if I'm not wrong, after this pr.
So, looking at its page we should add 
`juiz_sps_links`

to the disallowed classes.
Otherwise we'll give him a snippet :D

PR #193, opened 2015-06-24
fix: RTL based .row-fluid to be margin-left or margin-right

JTS-IL:
use margin-right or margin-left according to @is_rtl


Nikeo:
merged in the dev branch.
thanks for this contribution Yaacov :)

Issue #194, opened 2015-06-24
Display a back to top arrow on scroll not working in Customizr Pro 1.1.10 and Customizr 3.3.27


giorgioriccardi:
The "Display a back to top arrow on scroll" option is not triggering the class .show anymore in Customizr Pro 1.1.10 and Customizr 3.3.27

I'm not sure if the problem is here:
https://github.com/Nikeo/customizr/blob/01e34e16e76904b93d65bc87cbd6ea9600a4898c/inc/assets/js/parts/main.js
because when I scroll down on the page I get this log error:
TypeError: f.className.indexOf is not a function
.../wp-includes/js/wp-emoji-release.min.js?ver=4.2.2
Line 4
etc. etc.

Customizr is running on a fresh clean WP 4.2.2 version, nothing else is installed. I'll try to figure out tomorrow

eri-trabiccolo:
Hello Giorgio,
thanks for reporting this, yep back to top isn't displayed.
That error, anyway, shouldn't be related to the back to top arrow. It happens when you have featured pages with no images, so when you display the holder. Also it should happen with Customizr-Pro and not with Customizr 3.3.27 free.

giorgioriccardi:
I double checked and it was not working with 3.3.27!
Works into 3.3.28, obviously ;-)

eri-trabiccolo:
Hello giorgio:
"That error, anyway, shouldn't be related to the back to top arrow. It happens when you have featured pages with no images, so when you display the holder. Also it should happen with Customizr-Pro and not with Customizr 3.3.27 free."

I meant, the error you reported:
TypeError: f.className.indexOf is not a function
.../wp-includes/js/wp-emoji-release.min.js?ver=4.2.2

should not happen with 3.3.27.
And as I said was totally unrelated to the btt-arrow issue ;)

giorgioriccardi:
Got it, that happens when I try to read fast, I focus on the straw without seeing the tree

eri-trabiccolo:
:D 
You cannot read my English fast!!! :D

PR #195, opened 2015-06-25
Fix: re-introduce btt-arrow handling in new front js

eri-trabiccolo:
Back to top arrow js code was missing in the new front js, as reported in #194 
This PR reintroduces it.
I also thought to move the window's scroll event listening/handling in the user experience jq plugin, since both btt-arrow and sticky header share the same code.
Dunno what you think or whether you want to treat it in a different way, anyway.. 
bug fixed :P

### DIFF
--- a/inc/assets/js/parts/_main_sticky_header.part.js
+++ b/inc/assets/js/parts/_main_sticky_header.part.js
@@ -13,8 +13,6 @@ var czrapp = czrapp || {};
       this.elToHide         = []; //[ '.social-block' , '.site-description' ],
       this.customOffset     = +TCParams.stickyCustomOffset;
       this.logo             = 0 === this.$_sticky_logo.length ? { _logo: $('img:not(".sticky")', '.site-logo') , _ratio: '' }: false;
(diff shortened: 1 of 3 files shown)


Nikeo:
Hi Rocco, thanks for this !!

About the event listening, even if it is redundant, I will rather treat them separately in each subclass.

eri-trabiccolo:
Ok then I'll push a new pr, in which I'll anyway correct this:
https://github.com/Nikeo/customizr/blob/master/inc/assets/js/parts/_main_sticky_header.part.js#L73

which should be this.timer,
am I wrong?

Nikeo:
OK
yes I do think so

Nikeo:
I mean, you're not wrong :D

Nikeo:
Can you push it this afternoon ? I will try top update the theme today with this quite important fix.
Let me know if you can't, I'll do it then.
Thanks !

eri-trabiccolo:
I'm on it now,
it will be basically the same of this pr, but with separated scroll handlers

PR #198, opened 2015-06-27
Fix: better check on whether display or not attachment content

eri-trabiccolo:
I've found that there could be one case in which the theme tries to display the attachment anyways.
It happens, because of some plugin, when $post isn't set but is_singular() returns true .
Since the $post is still not set the only thing, in that case, that will be displayed is:
1) the title, if any, because of the __before_content action hook fired
2) the image navigation

Both unneeded :D

This will fix it.


Nikeo:
ok. thanks for catching this. Ready to merged

PR #200, opened 2015-06-28
Correction of few comments for the RIGHT sidebar

giorgioriccardi:
commented ##hook for the RIGHT sidebar
__after_article_container


Nikeo:
changed + pushed in dev.
Thanks!

PR #202, opened 2015-07-03
Fix: slider front and backend improvements

eri-trabiccolo:
1) Options added:
    - use a custom link
    - link whole slide
    - open link in a _blank target
  2) Further improvements:
    2a) back/admin
      - php
        - unify, where possible/needed, attachment and post metabox
          html structure
        - use minified scripts when WP_DEBUG == true
      - js
        - revamp tc-slider-ajax.js, better structure avoiding code
        duplication
    2b) front
      - php
        - Cache slider's model per slider
        - Store number or rendered sliders and use this value as part of
          the slider id
      - js
        - Cache sliders elements
        - Allow multiple sliders in the same page
        - Handle whole slide linking in mobiles ( suppress slide next on
          tap event )


eri-trabiccolo:
:( Comment indentation lost above.
Looks fine in the git commit log though

Nikeo:
Tested. Merged with minor changes.
Comments on asana.
Thanks !

eri-trabiccolo:
Fine,
thanks for the feedback too ;)

PR #203, opened 2015-07-03
Fix: do not try to access maybe not existent TC_featured_pages class

eri-trabiccolo:
Reported here:
http://presscustomizr.com/support-forums/topic/the-forms-of-plugin-formidable-have-stopped-working/

When previewing that form's plugin we're in admin context where front classes don't exist, then the fatal error.


Nikeo:
thanks :)

PR #204, opened 2015-07-04
Fix: slider-controls color when clicked

eri-trabiccolo:
Should fix the floowing issue:

When you click on an arrow, it keeps the hover color
even when the mouse is out.

p.s.
tested in chrome and ff
I think it's fine.


Nikeo:
cool, looks good and simple.
Thanks

PR #205, opened 2015-07-05
Fix: video post format, show full content in alternate layout

eri-trabiccolo:
https://wordpress.org/support/topic/videos-are-not-available-on-the-main-for-video-type-posts-any-more?replies=1
Mmm, why was that removed?

p.s.
I'll be away in 5minutes or so, for the rest of the day (pretty much)


Nikeo:
really have no clue why it's been removed...
thanks

PR #206, opened 2015-07-06
Fix: display slider-loading-gif only if js enabled

eri-trabiccolo:
Following to this ->
https://wordpress.org/support/topic/replace-slider-loadergif?replies=4#post-7128317

What do you think about the pure javascript solution in this PR?

### DIFF
--- a/inc/parts/class-content-slider.php
+++ b/inc/parts/class-content-slider.php
@@ -333,11 +333,16 @@ class TC_slider {
     if ( 'demo' != $slider_name_id && ( 1 != esc_attr( TC_utils::$inst->tc_opt( 'tc_display_slide_loader') ) || ! apply_filters( 'tc_display_slider_loader' , true ) ) )
       return;
     ?>
-      <div class="tc-slider-loader-wrapper">
+      <div id="tc-slider-loader-wrapper" class="tc-slider-loader-wrapper" style="display:none;">
         <div class="tc-img-gif-loader">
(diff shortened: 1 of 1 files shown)


Nikeo:
yes looks good like that.
I think we should just add a comment in the code to remember why we've added this line of code...since javascript is rarely disabled!

eri-trabiccolo:
Yeah I know but that user says otherwise:
https://wordpress.org/support/topic/menu-without-javascript?replies=1#post-7128266

:D 
Do you want an additional PR or will you take care about adding the comment?

Nikeo:
I'll add them :)

giorgioriccardi:
Actually I was surprised as well when I found out that several of the Oceania, Africa and South America countries are actually disabling JS on their browsers, well, in rural areas anyway.
Apparently is a internet speed/bandwidth matter; said that developing fallback solution is time consuming, but may be a general warning when js disabled is detected might alert people about the menu problem before developing all the site with Customizr...
but I guess that some sort of problem might rise with any modern WP theme ;-)

Nikeo:
yes modern user experience is totally js dependant
it might not be the case with future CSS enhancements...

giorgioriccardi:
what future CSS enhancements are you talking about? what do you know guys that I don't know about the future...?
I know only this https://make.wordpress.org/core/2015/07/06/singular-php-new-theme-template-in-wordpress-4-3/ and is not even a top discovery

Nikeo:
Hi giorgio, I don't think I know much more than you.
From what I understand, there's a constant trend to embed more js feature in the CSS like animations.

PR #207, opened 2015-07-09
Fix: do not try to enqueue non existent iphonecheck min js file

eri-trabiccolo:


Nikeo:
Fixed in a different way => the minified version will still be loaded if WP_DEBUG == false.
I've also fixed a bug in the way grunt "uglifies" this js.
thanks :)

eri-trabiccolo:
:)

PR #209, opened 2015-07-13
Fix: small css fixes for the new def menu

eri-trabiccolo:
For some reason I didn't include this in the first PR.
Mostly due to an unwanted undo.

Sorry


eri-trabiccolo:
Embedded in #210

PR #210, opened 2015-07-14
New menu fixes

eri-trabiccolo:
Following small fixes.
In each commit a summary of the changes.

We still have to decide the behavior of the header menu button-div in some circumstances:
1) sticky-enabled with no tagline (clear: both?)
2) no tagline (clear:both, and some margin?)
3) tagline but no socials (and logo on the left -> menu button on the right)
the third is pretty tricky, cause clear:both will place the menu-button below the tagline leaving a lot of useless space on the left

I don't know whether or not you want to merge this, but please take a look at the small fixes asap, otherwise I have to make another PR :D

### DIFF
--- a/inc/admin/js/parts/_control.js
+++ b/inc/admin/js/parts/_control.js
@@ -286,14 +286,24 @@
       }
     },
     'tc_menu_style' : {
-      controls: [
-        'tc_menu_type',
-        'tc_menu_submenu_fade_effect',
-        'tc_menu_submenu_item_move_effect',
-        'tc_menu_resp_dropdown_limit_to_viewport'
-      ],
-      callback: function (to) {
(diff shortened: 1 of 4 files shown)


Nikeo:
Fixed in dev.

eri-trabiccolo:
Mmm, but the margin top was just for the sidenav nav (so the menu container on the side nav)
#tc-sn .tc-sn-inner.nav-collapse .nav

so I don't get the relationship with the tagline :D

Nikeo:
the tagline stuff is a visual alignment issue I just noticed

eri-trabiccolo:
Ah fine, so ok, you don't like the sidenav nav margin-top anyway :D

PR #211, opened 2015-07-15
Fix: html5shiv for ie9-

eri-trabiccolo:
(*) remove shiv from modernizr (still loaded in the footer)
(*) add new js file html5.js
(*) use conditional html comment in the header to load the shiv
    just for ie9-


eri-trabiccolo:
I think it's fine .. what do you think?

Nikeo:
Hi Rocco, why this PR since it works now that modernizr is loaded in <head> for ie?
I'm sure there's something I miss !
please enlighten me ;) !

eri-trabiccolo:
:D Nothing special but the html5shiv is needed just for ie9- (am I wrong?) so why having it in modernizr? Which btw is loaded in the header also for other ie versions (am I wrong?) we save kBytes for modern browsers and load the shiv just for ie9- in the header. 
That's all :D

Nikeo:
the shiv is needed just for ie9- => yes mainly I agree but apparently it also applies basic styling for HTML5 elements for Safari 4.x and FF 3.x but I don't know what !! => those are not going to be applied anymore (safari 4.x is 5 years old and FF 3.x 4 years old...)

yes it's loaded in head even when not needed which is not optimal

that makes sense I agree, there's no small savings when it comes to speed :) and the Customizr theme is still far from being optimal from this standpoint. But this will change in V4! :)

Ok I merge this then thanks

eri-trabiccolo:
:)
Anyway Customizr loads pretty fast IMHO :D

Nikeo:
About speed, would it be absurd from your point of view to insert tc_common.min.css into the style.css of the theme and this way save a request ?

What's your opinion ?

eri-trabiccolo:
I read about this idea you had.
I cannot answer "would be a small saving" anymore. 
What I think is that, mm it's a "saving" (one request less) just for people who doesn't have a child theme. 'Cause in the latter case still 3 styles are loaded.
I also think that there are people who, sometimes, directly add the custom css to the main style.css
The change can produce the following:
1) user get scared about the amount of code there's in the style.css and will be make a child-theme => fine :D but then no save, 'cause still 3 files will be loaded
2) user will still append something in the style.css breaking something :D or some custom rules could be overridden by the skin and the custom css box's style. 
It will be a little mess, probably, in terms of priority, but maybe we can just make a choice and inform users of that :D 

That's my opinion :D

Nikeo:
yes, I agree, I also thought about the "scaring" amount of css and the risk that user mess with it.
we better "hide" the mess, sort of. :)

ok let's bury this idea

PR #212, opened 2015-07-15
Fix: typo when trying to animate html,body

eri-trabiccolo:
We never noticed this 'cause if a pretty rare circumstance.

p.s.
not related to the preview issue, which seems to be a wordpress issue


eri-trabiccolo:
included in #214

Nikeo:
already merged

PR #214, opened 2015-07-16
Fix: front-js, dropdown-to-viewport fix

eri-trabiccolo:
The new algorithm is based on the effective position of the expanding
element in the viewport, which is computed as the top offset of the
element minus the window scroll top.

Tested with my smartphone both when sticky-header on (scrolling or not)
and sticky-header off (scrolling or not)

p.s.
additionally a small typo fix, missing jquery alias ($) when trying
to animate html,body if stiky-header and no TCParams.dropdowntoViewport

### DIFF
--- a/inc/assets/js/parts/bootstrap.js
+++ b/inc/assets/js/parts/bootstrap.js
@@ -1656,8 +1656,7 @@ var TCParams = TCParams || {};
 
       if ( TCParams && 1 == TCParams.dropdowntoViewport )
       {
-        var tcVisible = $('body').hasClass('sticky-enabled') ? $(window).height() : ($(window).height() - $('.navbar-wrapper').offset().top);
-        tcVisible = ( tcVisible - 90 ) > 80 ? tcVisible - 90 : 300;
+        var tcVisible = czrapp.$_window.height() - this.$element.offset().top + czrapp.$_window.scrollTop();
         this.$element.css('max-height' , tcVisible + 'px');
       }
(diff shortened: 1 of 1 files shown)


Nikeo:
Hello Rocco, can you recall me what is the initial problem addressed here ?
thanks !

eri-trabiccolo:
The initial problem? Mmm just found this out while doing other stuff, but you can reproduce the issue on presscustomizr.com in mobile version, just shrink the page width to the mobile menu threshold, and shrink it vertically too, open the mobile menu, you'll see the nav-collapse overflows the window, so when sticky header you cannot really see the whole menu.
This is because (I think) of the hero unit above the header.
Other than the fact that I really like how clean is my solution :D

eri-trabiccolo:
p.s.
someone reported the issue here:
http://presscustomizr.com/best-web-designs-based-on-the-customizr-wordpress-theme-of-january-2014/comment-page-1/#comment-328491

It related his issue with the presscustomizr.com one, but they're not. He doesn't have dropdown to viewport checked, his issue might depend on some conflicting plugin, or a bug in his version (3.3.23). In his case the sticky-header should be disabled when the menu is open, but it doesn't happen, dunno why

Nikeo:
Ok thanks :)
yes looks clean, much cleaner than the previous one indeed !

eri-trabiccolo:
Thanks, I had too many rejected PR this week :P

PR #216, opened 2015-07-19
New sticky footer

eri-trabiccolo:
against c3e8833187


inc/assets/js/parts/_main_sticky_footer.part.js
```diff
@@ -0,0 +1,101 @@
...
+    },
+
+    /***********************************************
+    * DOM EVENT LISTENERS AND HANDLERS
+    ***********************************************/
+    stickyFooterEventListener : function() {
+      var self = this;
+
+      // maybe apply sticky footer on window resize
+      czrapp.$_window.on( 'tc-resize', function() {
```
eri-trabiccolo: this should be $_body too, but anyway it propagates so.. no problem
or maybe it should be fired on $_window in czrapp since it regards to the $(window)

eri-trabiccolo:
What happened to all my PR..
I think I have to do this again... let me know

PR #217, opened 2015-07-19
Fix: the cursor of the whole btn-toggle-nav div should be

eri-trabiccolo:
a pointer since the whole area is clickable.

Made just for the tc-side-menu btn because there will be no label for the responsive one (so regular menu isn't affected by this issue)


eri-trabiccolo:
mah, made some confusion with git.
anyway you already did this in the second menu branch.
Close this.

PR #219, opened 2015-07-20
Fix: correct mispelled "Design" translation in russian.

eri-trabiccolo:
According to Dmitry Pak there's no "дезайн" in russian.
http://presscustomizr.com/support-forums/topic/order-output-units/#post-86779

p.s.
I didn't touch the "fuzzy" translations affected by the same typo, 'cause they're still fuzzy.

against:
c3e88331878ac


Nikeo:
cool thanks

PR #220, opened 2015-07-21
Fix: avoid blocks reordering when they contain at least one iframe

eri-trabiccolo:
In truth this can be done better.
We can check the presence on just left and content and say:
- is there an iframe in the left sidebar? Move the content before the left sidebar
- is there an iframe in the content? Move the left sidebar after the content
( So keeping the interested one fixed (right sidebar should not be moved, so we can skip the check on it) )
- is there an iframe in both? Don't re-order.

I can do this if you agree, but I need a little of time (an afternoon), so let me know ;)

p.s.
also consider that the check is performed in the reordering method because on document ready the twitter timeline isn't an iframe yet, so if I do this in the dynSidebarReorder It will not see twitter timeline iframe.

### DIFF
--- a/inc/assets/js/parts/_main_userxp.part.js
+++ b/inc/assets/js/parts/_main_userxp.part.js
@@ -191,10 +191,13 @@ var czrapp = czrapp || {};
       that.$_left         = that.$_left || $("#main-wrapper .container " + LeftSidebarClass);
       that.$_right        = that.$_right || $("#main-wrapper .container " + RightSidebarClass);
 
+      // do nothing if there's at least one iframe
+      if ( that._has_iframe( [this.$_content, this.$_left, this.$_right] ) ) 
+        return;
+
       //15 pixels adjustement to avoid replacement before real responsive width
       switch ( _sidebarLayout ) {
         case 'normal' :
-        console.log('JOIE DU NORMAL');
           if ( that.$_left.length ) {
             that.$_left.detach();
             that.$_content.detach();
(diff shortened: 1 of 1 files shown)


eri-trabiccolo:
Nico,
just seen we have some issues with this.
I was doing another version, the clever one. But we still have an issue, when both containers has an iframe... and one of them is the twitter timeline.

Nikeo:
can you push a version where we check if there's at least one iframe in any container ?
thanks Rocco :)

Nikeo:
Another topic quick question : about the sticky footer, why don't we enable it by default ? Any possible side effect ? what do you think ?

eri-trabiccolo:
Possible side effects I can think about:
1) A lot of people already have some alternative (not very good) way to achieve the same, see my snippet
2) This one is not perfect because it reacts to window's resize not to the content resize. I have to see if I can with some observer (but I tried, looks like I'm not able to listen to an element height change). Example:
think about photon which doesn't reserve an eight to the images (no attrs) the content height increases (more evident when img-smartload is enabled) but the sticky footer doesn't react (I was just asking you if you thought we have make it react to 'tc-smartload' event too, I think so, at least when they use our smartload will be fine). This means that in some edge-cases when you have a small content with images not loaded (see photon) the footer will be pushed down. Then when the images will be fully loaded the content will grow but the "footer push" will still stay the same.
So => not perfect at all

Nikeo:
ok
let's "observe" how users will react to it for a period then and leave it disable by default
I'm pretty confident but we'll see.
thanks

eri-trabiccolo:
I'm done with the "clever version" (no need to check the right, but the concept is the same)

PR #226, opened 2015-07-27
Fix: anchors smooth scroll - exclude ultimate members anchor links

eri-trabiccolo:
reported here:
http://presscustomizr.com/support-forums/topic/ulitmate-member-plug-in-problems-with-customzr-pro


Nikeo:
Thanks. :) I've added the following note below this option in the customizer : 
this option can create conflicts with some plugins, make sure that your plugins features (if any) are working fine after enabling this option.
Too many problems with this option :/

eri-trabiccolo:
Yeah, hopefully it will be read

Nikeo:
yes hope so :)

PR #230, opened 2015-08-06
Fix: do not update the tb post meta as soon as you enter tha add post

eri-trabiccolo:
page

p.s.
This is more a reminder.
Fixes the issue reported here: http://presscustomizr.com/support-forums/topic/conflict-with-press-permit-core/
Basically `save_post` is fired also when you enter the add new post page.
This is mainly a plugin issue which we facilitated.
Anyway there's no reason, imho, to add an empty post meta in that case.

The same thing should be done in fpu too (let me know if you want I make a PR also there or you'll do it manually :page_facing_up: )


eri-trabiccolo:
Do you want a PR for fpu or you'll do it?

PR #232, opened 2015-08-10
Fix: btn-toggle-nav btn box-sizing must be content-box

eri-trabiccolo:


Nikeo:
Hi, can you quickly remind me what's the problem solved here ? (totally forgot about this... :|)

eri-trabiccolo:
Well not reported anywhere.
It's just that we have:
-webkit-box-sizing: border-box;
-moz-box-sizing: border-box;
box-sizing: content-box;

It should be content-box for every browser. Or not?

Nikeo:
Humm not sure. Don't know very well those properties
need to check here https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing

Nikeo:
but it must be consistent, either one or the other of course

eri-trabiccolo:
Well, the thing is.
When you set the width/height of an element by default you set the width/height of its content, padding and borders excluded (content-box).
Setting the box-sizing to "border-box" means that the "width/height" property refers to the whole element, padding and borders included but margin excluded.

I needed it to be a content-box because at first the button had some padding (while rotating the icon bars the height grown.. remember that?), now it hasn't, so we could even get rid of those rules. But I think will be better keep the content-box

Nikeo:
ok make sense now. Let's merge it then, even if not really needed now as you said.
Thanks !

PR #233, opened 2015-08-12
Fix: compatibility fixes with old wp version

eri-trabiccolo:
1) customize: _control.js
old wp versions don't have wp.customize.section (see gc pull #4)
same method

2) customize-preview: _main_base.part.js
when in customize and preview loaded I got this error:
Uncaught TypeError: Cannot convert undefined or null to object
on line:
https://github.com/Nikeo/customizr/blob/master/inc/assets/js/parts/_main_base.part.js#L218

and most of the preview js break

the simple change in this pr will fix it

3) customize: class-fire-utils_settings_map.php
Simple wrong strong tag closing. This made the the preview iframe
overlay the customize settings (all the preview in the left content), due to html breakage

4) class-header-nav_walker.php
Avoid using wp filter nav_menu_css_class which could cause api breakage
with old wp versions, and most likely, some plugin.
Replace it with a custom filter hook tc_nav_menu_css_class we fire over
the element's classes in our nav_walker display_element override.

To me these simple changes sound good and should not cause any issue,
but just take a second look on them ;)

### DIFF
--- a/inc/admin/js/parts/_control.js
+++ b/inc/admin/js/parts/_control.js
@@ -4,7 +4,8 @@
  * @since Customizr 1.0
  */
 (function (wp, $, _) {
-  var api = wp.customize;
+  var api = wp.customize,
+      $_nav_section_container;
 
   /**
    * @constructor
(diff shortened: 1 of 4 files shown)


Nikeo:
really clean stuffs, cool

eri-trabiccolo:
:) Yeah, the filter in the nav walker isn't really needed atm, I just left it as filter for readability and modularity.. I tend to make a lot of functions too :D

PR #234, opened 2015-08-12
Add: new plugin compatibility - BuddyPress Global Search

eri-trabiccolo:
https://wordpress.org/plugins/buddypress-global-search/

"issue" reported here:
https://wordpress.org/support/topic/plugin-theme-conflict


eri-trabiccolo:
Needs more tests

PR #235, opened 2015-08-14
Fix: properly remove box around navbar when no menu is set

eri-trabiccolo:
Bug reported here:
https://wordpress.org/support/topic/dispay-menu-in-a-box?replies=6

Basically when no menu is set body class to not display the box is not
added because the filter callback is set just when menu are shown in the
class-header-menu.php

Moving the class adding to class-header-menu.php, which I think is a
more proper place for it 'cause it regards the whole navbar-wrapper more
than menus, solves the issue.


Nikeo:
yep 100% agree. Thanks

PR #239, opened 2015-08-22
Fix: hide menu controls on nav section expanded only for newer wp

eri-trabiccolo:
versions avoiding conflicts with old wp versions

Basically we hide them just when api.section is a function, 'cause we don't need that for old wp versions which also don't have that function.
Just a question, did you test the new code with 4.1<=wp<4.3 ?


Nikeo:
handled differently locally.

Nikeo:
pushing the code for a build

eri-trabiccolo:
Ok good ;)

PR #243, opened 2015-08-24
Fix: qTranslate-X compat code improved

eri-trabiccolo:
1) Fix an issue regarding to qtranslate-X filtering theme options and
overwriting some multilanguage theme's options in the db with the
current language translation of them
2) Get rid of outdated code regarding fp-text
- we don't need to remove the char limit because we apply the
  translation before trimming the text
- we don't need to force the use of the page excerpt when qTranslate is
  enabled 'cause we already to that in any case

### DIFF
--- a/inc/class-fire-plugins_compat.php
+++ b/inc/class-fire-plugins_compat.php
@@ -15,10 +15,13 @@ if ( ! class_exists( 'TC_plugins_compat' ) ) :
     //Access any method or var of the class with classname::$instance -> var or method():
     static $instance;
     function __construct () {
+        
       self::$instance =& $this;
       //add various plugins compatibilty (Jetpack, Bbpress, Qtranslate, Woocommerce, The Event Calendar ...)
       add_action ('after_setup_theme'          , array( $this , 'tc_set_plugins_supported'), 20 );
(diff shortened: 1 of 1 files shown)


eri-trabiccolo:
Do we want to remove also this block?
        /* The following is pretty useless at the moment since we should inhibit preview js code */

It's really useless, I'll comment it for the fpu, 'cause it doesn't work very well, basically in the preview the fp-text isn't translated anyway and in fpu change the transport causes, then, just an useless refresh when typing.

And in Customizr that filter looks like it doesn't do anything, probably because of theme's api changed.

Nikeo:
which block Rocco, I don't get it

Nikeo:
the comment ?

eri-trabiccolo:
From https://github.com/eri-trabiccolo/customizr/blob/master/inc/class-fire-plugins_compat.php#L221
to 224

Nikeo:
ah ok yep, for next release then :)

eri-trabiccolo:
yes, fine ;)
Basically are the same changes made to the core theme + those commented lines which are more a limit than a feature :)

Issue #258, opened 2015-08-26
Product tabs not showing in Woocommerce if “Smooth scroll on click” is enabled.


Nikeo:
The link smooth scroll should be disabled by default (if not the case already...)

eri-trabiccolo:
fixed in af83b9bff08b

PR #259, opened 2015-08-26
Fix: disable link smoothscroll in woocommerce contexts

eri-trabiccolo:
See issue #258


eri-trabiccolo:
Otherwise, we should implement an "exclude selectors" which isn't based only on the link CSS class but also on the parents classes/id, exactly as we do with the external links.
https://github.com/Nikeo/customizr/blob/master/inc/assets/js/parts/jqueryextLinks.js#L73

this method could be ported in czrapp constituting an utility.
What do you think?

Nikeo:
yep agreed 100% !

eri-trabiccolo:
Perfect, will do that at my return ;)

eri-trabiccolo:
so you chose this one?

PR #155, opened 2015-05-08
UPDATE : RTL-ing main elements

JTS-IL:
A very long walk-trough /rtl/tc_common.css with small adjustments in some php code files that now support both RTL and LTR sites without too much css tricks

PR #263, opened 2015-08-27
Fix: rtl grid fixes - margin property rtl dependent

eri-trabiccolo:


JTS-IL:
Interesting solution, but as i remeber, the left/right margin is not always as the text flow direction.
I have a dev site, how can I test this fix?

JTS-IL:
I need some help here.
I can't understand where @floatdirection is changed from left to right for RTL

eri-trabiccolo:
what do you mean? floatDirection has been introduced by Nico, after this PR->
https://github.com/Nikeo/customizr/pull/155/files
I think

JTS-IL:
155 was my major RTL project on customizr, I don't see @floatdirection on #155
I don't think I saw @floatDirection before (maybe I'm old and my memory is getting rusty...)

JTS-IL:
Well, your hint lead me here:
https://github.com/Nikeo/customizr/commit/0765878b50d376cb6e713cbca11619f17a512533

eri-trabiccolo:
Hello,
sorry if I've been not clear.
I didn't say you did it or that it was there.
I said: 
>floatDirection has been introduced by Nico, **after** this PR->
>https://github.com/Nikeo/customizr/pull/155/files
>I think

JTS-IL:
I'm definitely getting old :-)
Didn't see the "after"  he...he...he...

eri-trabiccolo:
eheh :D

eri-trabiccolo:
I forced a new version of this PR.
In my idea RTL (front) should be mostly handled with CSS.
There are still things to do, but I think it could be a good "starting point".
It also removes direction dependent rules from the skins.
I haven't touched the thumbnails for now.
Also menu and logo, since there's an option to put them on the right or on the left and left and right are universal concepts, the RTL version still respects the options.
Regarding this, sidebars make an exception, 'cause I think (as JTS would hopefully confirm), that left and right sidebars are more seen as "First" and "Last", so it makes sense that the left sidebar shows up on the right (it's already so in the current RTL version - implemented via PHP), for RTL users - 'cause it generally contains widgets you want to see "before" the content.
Would like to know JTS point of view (you can test the "compiled" version of this branch if you don't want to mess with grunt :D -> https://github.com/eri-trabiccolo/customizr/tree/rtl-fixes-build )

p.s.
I also followed JTS suggestion about having a counterFloatDirection , though I think that now we can easily use just "direction" and "counterDirection" instead (also getting rid of textAlign, replaced by "direction").

JTS-IL:
Hi, 
I've been on a vacation (not far from you I think...) and began reading all my eMails just today...
About left and right sidebars - your suggestion of "first" and "last" is exactly what I was thinking during the vacation. I'm working on a multilingual site with polylang (both RTL and LTR) and had some issues with the positioning of logo and sidebars, so I suggest to make it even more customizable - define if the position is absolute (Right or Left) or relative (directional or counter-directional)
I believe this setup will bring many users that need a bi-directional multi-language support to use the theme (and the PRO version).
I know the Hebrew sites often have a "mirrored" English version, I think that many Arabic sites also have a French or English version.

JTS-IL:
Is it possible to merge #273 into this branch?

JTS-IL:
Well...
It is a good (very good) starting point.
As I wrote earlier - Logo and menu are also usually positioned as a "mirror image" when having a bidirectional site.
The RTL-FIXES-BUILD is installed here : jts.glezer.co.il, so you can see that the ENGLISH version looks a little strange.

eri-trabiccolo:
Another commit (also in the build, so you can test it easily).
I see your point about the mirroring. What I think is that despite you're obviously true, the header layout is still a design choice. There are LTR users who prefer have the logo on the right, maybe because they're left-handed, who knows :D 
Anyway we can:
1) force the mirroring
(you can use something like this, for the moment in your child-theme functions.php -> 
https://gist.github.com/eri-trabiccolo/a81345472c6ee3414886
just for the front, the live preview still needs work, but anyways...
)
2) add other 3 options for menus + logo left/right specific for RTL.
The ball passes to Nico :D

JTS-IL:
Option 1 is very interesting solution.
I will test it tomorrow more thoroughly.

JTS-IL:
Hi, 
I activated the new build on another site : www.litvaks.org.il, added the my_rtl_layout function to functions.php, but the logo keeps staying on the selected side (left or right) for both directions.
what did I do wrong?
Any suggestion?

Issue #262, opened 2015-08-27
RTL : The site header image jumped above the main menu


Nikeo:
![niq9laotzmxnziyiynrte-r8pymh1df7oay88kmyesi](https://cloud.githubusercontent.com/assets/3038156/9514889/e8188e18-4c9d-11e5-85cf-58f38974afbe.png)

JTS-IL:
?
Looked at his site, it looks great.
Did he fix something?

Nikeo:
he downgraded to 3.4.28 because he had a serie of issues related to RTL, this one and also #260 and #261

JTS-IL:
OK!
digging...

JTS-IL:
The issue is in mixins.less
I'm having some problems with github, so here is the fix:
old code:
***********************
    .row-fluid {
      width: 100%;
      .clearfix();
      [class*="span"] {
        .input-block-level();
        float: @floatDirection;
        & when (@is_rtl = false) {
        margin-left: @fluidGridGutterWidth;
        *margin-left: @fluidGridGutterWidth - (.5 / @gridRowWidth * 100 * 1%);
        }
        & when (@is_rtl= true) {
        margin-right: @fluidGridGutterWidth;
        *margin-right: @fluidGridGutterWidth - (.5 / @gridRowWidth * 100 * 1%);	
        }
      }
      [class*="span"]:first-child {
        & when (@is_rtl = false) {
        margin-left: 0;
        }
        & when (@is_rtl = true) {
        margin-right: 0;
        }
      }
***********************
new code
***********************
    .row-fluid {
      width: 100%;
      .clearfix();
      [class*="span"] {
        .input-block-level();
        float: @floatDirection;
        & when (@is_rtl = false) {
        margin-left: @fluidGridGutterWidth;
        *margin-left: @fluidGridGutterWidth - (.5 / @gridRowWidth * 100 * 1%);
       margin-right: 0;
        }
        & when (@is_rtl= true) {
        margin-right: @fluidGridGutterWidth;
        *margin-right: @fluidGridGutterWidth - (.5 / @gridRowWidth * 100 * 1%);
        margin-left: 0;	
        }
      }
      [class*="span"]:first-child {
        margin-left: 0;
        margin-right: 0;
        }
      }
***********************
that should fix #261 and #262

eri-trabiccolo:
Hello JTS,
though it works, I'm not totally sure this is the dual of the ltr rule.
Would you like to test this PR -> #263
[...]

PR #265, opened 2015-08-27
Update grid.less - fix logo position (#262)

JTS-IL:
Accepting Rocco's suggestion to add a rule for pull-left


Nikeo:
merged in dev

PR #278, opened 2015-09-08
Fix: expand last published sticky post in the grid

### LINKED ISSUE #277: Grid expanded post
At the moment the expanded sticky post is not the latest published but just the first in the array of the sticky_posts option. We should retrieve the latest published among those posts.A custom SQL should be the less expensive way.

eri-trabiccolo:
Fixes #277

### DIFF
--- a/inc/parts/class-content-post_list_grid.php
+++ b/inc/parts/class-content-post_list_grid.php
@@ -830,7 +830,7 @@ if ( ! class_exists( 'TC_post_list_grid' ) ) :
         * check if we have to expand the first sticky post
         */
         private function tc_is_sticky_expanded( $query = null ){
-          global $wp_query;
+          global $wp_query, $wpdb;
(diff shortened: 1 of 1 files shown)


Nikeo:
merged in dev

eri-trabiccolo:
Anyway I missed this comment:

> Now it also set the "expanded_sticky" property.

it already did that (I remember that at a certain point we introduced that), the only difference is that now it does it better.

eri-trabiccolo:
Here's where we (I) introduced the caching:
https://github.com/Nikeo/customizr/commit/506e9449587b2e2bc9a76a292f0459603503d413

eri-trabiccolo:
Actually this PR misses a condition in the query:
AND post_status = 'publish'

I'll propose a PR to fix this + new function to cache the sticky post, in a few hours (lunch time here ;))

PR #290, opened 2015-09-14
Fix: Woccomerce's product tabs not showing if Smooth scroll on click …

eri-trabiccolo:
…enabled

Fixes issue #258
This PR basically reproduce the code used in jqueryextLinks plugin to
check wheter or not a selector is allowed in the czrapp base code.
The new Smooth Scroll on click will work on:
a) (preliminary exclusion)
all the links which start with "#", except the ones specified in the
localized param anchorSmoothScrollExclude.simple (using jQuery.not)
b) (deep exclusion)
all the links which passed the preliminary exclusion and which do not
match the deep exclusion clauses on id and class - specified in the
param anchorSmoothScrollExclude.deep - using the method isSelectorAllowed.
The method isSelectorAllowed allow us to filter all the elements with a
parent matching the exclusion clauses on id and class.

This way we can exclude woocommerce's product tabs children of an
element with class 'wc-tabs'

p.s.
Since jqueryextLinks.js is a more general library (we can consider it external) 
I left the original code there with a call to the new czrapp method when czrapp 
exists, more to leave to you the decision on whether or not you want to 
remove the old code (to save some byte in tc-scripts.js)

### DIFF
--- a/inc/assets/js/parts/_main_base.part.js
+++ b/inc/assets/js/parts/_main_base.part.js
@@ -135,6 +135,31 @@ var czrapp = czrapp || {};
     },
 
 
+    //@return bool
+    isSelectorAllowed : function( $_el, skip_selectors, requested_sel_type ) {
+      var sel_type = 'ids' == requested_sel_type ? 'id' : 'class',
+      _selsToSkip   = skip_selectors[requested_sel_type];
+
+      //check if option is well formed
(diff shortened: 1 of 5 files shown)


Nikeo:
OK let's do that.
I'll merge this PR then

eri-trabiccolo:
Ok so once merged, I'll remove the code to disable the link smooth scroll on click option in woocommerce contexts, replacing it with a filter on the localized params.
I'll look at EDD too if you want, just let me know ;)

PR #296, opened 2015-09-15
Fix: allow customizr framework to generate addons default options

eri-trabiccolo:
- in class-fire-utils:
  - declare an array attribute, filterable, which
  stores the allowed prefixes ('tc_' for common customizr options)
  - new public method tc_is_customizr_option will check wheter or not
    the option with the passed key should be taken in account. The
    return of this method is filtered to allow addons using a different
    way to check the eligibility
  - tc_generate_default_options will use the above method to generate
    default options for the allowed keys
- in class-admin-customize the method tc_customize_factory will use the
  TC_utils method tc_is_customizr_option to check whether or not the
  option needs to be prefixed with the theme option group prefix


Nikeo:
excellent PR Rocco thanks !

eri-trabiccolo:
:D

PR #297, opened 2015-09-16
Fix: better handling of the expanded featured post in post list grid

eri-trabiccolo:
To actually know whether or not we have to expand the
sticky we need to know if there's a sticky post to expand. Sticky posts
option also stores trashed (not completely deleted) posts, so we have to
check that the post is 'publish'. So we can cycle among sticky posts
and check their status ( get_post_status( $ID) ). If we get at least one
publish post then we have a sticky expanded (always if we want to show
it).
The thing is that later on we need to use the latest sticky so we "need"
a custom query to order the publish sticky posts by date. We can do that
later or save time/resources using just one query which makes both
(avoiding get_post_status). And once we decide to do that why don't
cache the sticky_post ID (if any)?
So I'm with you, this function does more that just returning a
boolean, but I think is worthwhile. Anyway starting from this, if you
want to not cache the sticky post in this function we can remove the
caching line and do the cache elsewhere.
But remember that before caching the sticky post ID (or false) we cached
the boolean "has_expanded_sticky". So the actual difference is that
before we cached a bool, now we cache a tri-state which is useful in
other parts of the code. :)
My 2 cents (can you send me the invoice? :P).

### DIFF
--- a/inc/parts/class-content-post_list_grid.php
+++ b/inc/parts/class-content-post_list_grid.php
@@ -19,7 +19,6 @@ if ( ! class_exists( 'TC_post_list_grid' ) ) :
 
         function __construct () {
           self::$instance =& $this;
-          $this -> expanded_sticky = null;
 
           add_action ( 'pre_get_posts'              , array( $this , 'tc_maybe_excl_first_sticky') );
           add_action ( 'wp_head'                    , array( $this , 'tc_set_grid_hooks') );
@@ -826,12 +825,21 @@ if ( ! class_exists( 'TC_post_list_grid' ) ) :
         }
 
 
+
         /*
         * @return bool
         * check if we have to expand the first sticky post
+        *
+        * caches its ID or false
         */
         private function tc_is_sticky_expanded( $query = null ){
-          global $wp_query, $wpdb;
+          $_expand_feat_post_opt = apply_filters( 'tc_grid_expand_featured', esc_attr( TC_utils::$inst->tc_opt( 'tc_grid_expand_featured') ) );
+          if ( ! $_expand_feat_post_opt )
+            return false;
+          if ( isset( $this -> expanded_sticky ) )
+            return $this -> expanded_sticky ? true : false;
+          
+          global $wp_query;
           $query = ( $query ) ? $query : $wp_query;
 
           if ( ! $query->is_main_query() )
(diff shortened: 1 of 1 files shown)


Nikeo:
ok....
I know it's a pain, and there are other parts in the code not clean yet, but I would prefer to separate the states and the "model" (even if it's a model as simple as an id integer...)
Do you think there might be a way to do that ? Even if we need to add methods...

In its actual state, this code works fine and we understand it, but in the future, it might be hard to maintain if we keep this pasta salad of boolean and other data....:)

eri-trabiccolo:
Well we have to opt for a redundant code then.
The thing is, as I wrote above, that to actually know whether or not we have to expand a post we have to know that this post is sticky and publish, so the only way I can think about it (without involving an sql query in that phase) is with the procedure described above (check the post status of the sticky posts).
Then return as soon as we find a publish sticky post.
(wp get_post caches the post so actually shouldn't be a real issue).

But honestly the thing is that I called this 
tc_has_eligible_expanded()
what if I called it
tc_get_eligible_expanded_id()
and it returned the ID of the expanded or NULL?
And the check is on is_null ?
So basically it will be like "Hey give me the id of the eligible expanded" .. "Ah it's NULL, so no expandable posts".
I mean, get_var() can return the var requested (string) or NULL, so we can return ID (integer) or NULL :D
The whole problem, probably I missed this, is not this function right?
You mean that this field annoys you:
$this -> expanded_sticky

because is a tri-state. Because it can be NULL (not set), Integer (ID), FALSE (nothing to expand).
But it's a way to don't retrieve N (number of calls to tc_is_sticky_expanded() ) times the sticky posts then check which is eligible then retrieve the post id. 
I mean if we don't do that caching there we'll do that in another part, but still the tri-state is something we can use. Think about strpos : 

>This function may return Boolean FALSE, but may also return a non-Boolean value which evaluates to FALSE. 

Otherwise we need two vars, one stores the ID or NULL and another one stores if we already retrieved the expandable posts or not.

So the thing is,
we can even change that function (has_eligible_expanded) so to have a boolean return value, or change its name so it will suggest we can get an ID or NULL but still we need a tri-state to know whether or not we have to look for the existence of this post (which involves some "expensive" request) or we already did that and we found there were no suitable posts.
That's all the problem, basically I use the FALSE just to differentiate it from NULL which means that we haven't looked for it.
We can use an INT tristate 0 = not checked -1 = not found ID = found.

I can find a different solution of course, maybe a little bit redundant, but are you sure you'll like it?:D

About the model separation:
Don't forget we build a whole slider model basically to check whether or not we have the slider in a page :P 
This is the same thing, we try to get the ID of the post to expand and if no post we don't expand anything, I just wanted a way to not do this operation again. :)

eri-trabiccolo:
Mmm, you're probably thinking about having many grids in a page. Is this what I'm missing?
So standing the above comment (tri-state stuff).
What we have to avoid is 
https://github.com/eri-trabiccolo/customizr/blob/grid-better-sticky/inc/parts/class-content-post_list_grid.php#L839

before the checks on the wp query.
(?)

eri-trabiccolo:
Hello,
look that I've found that we can avoid all that stuff retrieving the sticky posts and the query altering.
So we can avoid caching the id.
Though I'm a little bit unsure on how to give a true sense at the code :D 
I'll think about it a little more, then I'll submit you another PR to let you see what I'm talking about :D

PR #298, opened 2015-09-17
Imp: link smooth scroll, exclude wc-tabs only in woocommerce contexts

eri-trabiccolo:
Is this what you meant?
Or did you mean to completely get rid of the deep param and build it up only when needed?
Let me know ;)


eri-trabiccolo:
Ah you probably should update travis settings imho after the repos migration.
Or, most likely, presscustomizr needs a travis account, right?

eri-trabiccolo:
See #301 
#302

I think both do the job fine, what do you think? What do you prefer?
Incremental(302) is more pretty, but 301 is "better" (less comparisons)

PR #299, opened 2015-09-17
Fix: remove now pretty useless classic emoji patch

eri-trabiccolo:
We introduced this because of a bug in 4<wp<4.2.3
fixed in 4.2.3 https://core.trac.wordpress.org/ticket/32305
which generated a js error when featured pages holders were displayed
(conflict with the svg element created by holder.js).
It was anyway a minor issue which we patched just to avoid issues
passing the theme review, because no production website actually
displays holders. Also it didn't break our js.
I don't think the, even minimal, overhead introduced by the previous
patch is still to be considered justified.
IMHO :D 

Do you agree?


Nikeo:
Yep 100%

PR #310, opened 2015-09-22
Fix: avoid navbar-wrapper overlaying title/logo in mobiles when no

### LINKED ISSUE #309: Navbar wrapper overlays logo/title when no site-description is displayed w<979px
I don't know when we introduced this bug :(

something like that should be fine.

`
@media (max-width: 979px){
  .tc-no-sticky-header .navbar-wrapper, .sticky-disabled .navbar-wrapper {
    clear: both;
  }
}
`

eri-trabiccolo:
site-description is shown

fixes #309

### DIFF
--- a/inc/assets/less/tc_custom_responsive.less
+++ b/inc/assets/less/tc_custom_responsive.less
@@ -135,10 +135,6 @@
     width: 100%;
     margin: 0%;
   }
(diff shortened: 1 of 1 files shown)


eri-trabiccolo:
did you merge both then? :D

PR #323, opened 2015-09-26
better translation

JTS-IL:


Nikeo:
Merged in dev. thanks !

PR #329, opened 2015-09-29
Fix: category-picker front do not query for non existent categories

eri-trabiccolo:


Nikeo:
Can you also add an option update on term deletion ?

eri-trabiccolo:
:)
I thought you wanted that.
Do you think is useful having both?
I mean, the "honerous" part is the front one, so I think that if we decide for the reaction on deletion we can avoid the front one.
Anyway I'll add the "option update" to this PR, then we'll decide whether or not having both is useful. :D

Nikeo:
yes I prefer to have both :)

eri-trabiccolo:
Done,
I "wrote" two ways to update the option, you can chose the one you prefer.
If you don't like any of them I will cry :P

Nikeo:
ah ah looks perfect Rocco. Put a smile on your face :)

eri-trabiccolo:
Thanks $GOD :D

PR #344, opened 2015-10-07
Fix: fix preview settings bind for wp versions <4.1

### LINKED ISSUE #327: WP 3.7.4 Customize Preview: Does the live preview (transport:postMessage) correctly work?
Need to investigate whether is a problem in the preview js or on how we change the setting value (with select2.val())

Actually here no options with transport postMessage are working with 3.7.4.
Probably an issue on my end then. Will look further more into it.

eri-trabiccolo:
fixes #327
I went for a conservative way, we'll see if in the future we might need more info on the wp version.


Nikeo:
Fine thanks.
I've just modified the param to make it feature aware instead of wp version aware.

PR #360, opened 2015-10-14
fix: accessibility compatibility

JTS-IL:
According to accessibility Guidelines: WCAG 2.0 (Level AA) instead of using b (bold) element,  "em" or "strong" should be used


Nikeo:
looks good thanks !

Nikeo:
merged in dev

PR #365, opened 2015-10-16
Fix: add third param for preg_match_all (required by PHP<5.4.0)

eri-trabiccolo:
Anyway most likely we have to change this way to process multiple bool conditions, 'cause we process some of them even when not needed.
Ex:
`a || b`
If a is true the expression above exits with true without evaluating b
But 
$cond = array(a,b); (bool)array_sum($cond); 
b is always evaluated.

Right?

p.s.
push the master :P

### DIFF
--- a/inc/class-fire-placeholders.php
+++ b/inc/class-fire-placeholders.php
@@ -285,7 +285,7 @@ if ( ! class_exists( 'TC_placeholders' ) ) :
         ! is_user_logged_in() || ! current_user_can('edit_theme_options'),
         ! self::$instance -> tc_is_front_help_enabled(),
         'disabled' == get_transient("tc_img_smartload_help"),
-        apply_filters('tc_img_smartload_help_n_images', 2 ) > preg_match_all( '/(<img[^>]+>)/i', $text ),
+        apply_filters('tc_img_smartload_help_n_images', 2 ) > preg_match_all( '/(<img[^>]+>)/i', $text, $matches ),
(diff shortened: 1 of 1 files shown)


Nikeo:
yes that's the caveat of this boolean processing

Nikeo:
If possible, I think we should develop in a php environment with older versions...
this happens too often I agree

eri-trabiccolo:
It happened mostly before, now we take care of that. I often look for the comp with old versions, this time I missed it :) 

Mmm I have to find a way to do that on my little linux machine :)
but which is the minimum PHP version we support?
Also:
https://wordpress.org/about/requirements/
suggested 5.6.0 but it also works with 5.2.4+
(so minimum 5.2.4?)

I think, overall, we have to start the process of supporting less wordpress versions, hence less php versions.. I think we have code compatibility for PHP-4 it's a pain in the .. :) but more than this we will never take advantage of the new PHP devs.

p.s.
some days ago a friend of mine had an issue with json_encode why?
Because of this:
https://github.com/WordPress/WordPress/blob/4.3.1/wp-includes/functions.php#L2674
(similar to our issue then)
The thing was that PHP declares the second param has been added in 5.3 hence that check. The host of this friend of mine has PHP 5.3.21 so it passed the check, still its json_encode prototype didn't have the second param.
I don't know whether it was a specific issue with that PHP version or the host used a buggy PHP or PHP doc is wrong... and so on :D

PR #378, opened 2015-11-02
Fix: better handling of the simple-load event triggered on holders

### LINKED ISSUE #377: jquery: Simple load function doesn't work for the holders
fp holders cannot be simple-loaded because they do not have a source (?).

Looking for the reason..

eri-trabiccolo:
fixes #377

### DIFF
--- a/inc/assets/js/parts/_main_jquery_plugins.part.js
+++ b/inc/assets/js/parts/_main_jquery_plugins.part.js
@@ -22,7 +22,7 @@ var czrapp = czrapp || {};
           $_to_center      = $( _where ).find('img');
 
       if ( smartLoadEnabled ) {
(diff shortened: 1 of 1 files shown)


Nikeo:
Ok thanks for this fix. Let's implement it for now. We'll see if we can improve this in the future.

eri-trabiccolo:
Probably the better thing will be fire the simple-load event inside the holder.js, if we want to do such a thing.

Nikeo:
humm, this would implied to maintain a modified version of holder, which I'm not really keen to do...

eri-trabiccolo:
Yes, that's why I went for this ugly hack :D

Nikeo:
not that ugly Rocco ;)

eri-trabiccolo:
:)

PR #385, opened 2015-11-04
Fix: to not match grid context when in search with no results

eri-trabiccolo:


Nikeo:
Cool Thanks.

Was this reported in the forum or did you find it by mistake?

Please keep describing your PR !

eri-trabiccolo:
Eheh, thought it was self descriptive.
Anyway I found, when doing other things (wpml compat) that when displaying the no result page I had:
1) the normal no result appearnce Title + no result content
+
2) no result content, and even if not visible also a grid element (a post) + all the classes added like tc-grid etc etc.

I'm not very able to totally reproduce it now but:
http://demo.presscustomizr.com/?s=aeraert

look at the source code you'll see:
`<div id="content" class="span12 article-container tc-post-list-grid tc-grid-shadow tc-grid-border tc-gc gc-title-dark-bg gc-white-title-hover gc-white-title tc-gallery-style">`

Make the same test with the PR applied and you'll see that those classes are not there anymore.

eri-trabiccolo:
Ops,
I see this 'causes a weird issue here, like infinite search results ..

PR #387, opened 2015-11-08
Add: WPML compatibility

eri-trabiccolo:
TODO: need to understand what to do with sliders of images, not clear
for now. Wait for WPML staff reply

### DIFF
--- a/inc/class-fire-plugins_compat.php
+++ b/inc/class-fire-plugins_compat.php
@@ -14,6 +14,9 @@ if ( ! class_exists( 'TC_plugins_compat' ) ) :
   class TC_plugins_compat {
     //Access any method or var of the class with classname::$instance -> var or method():
     static $instance;
+    //credits @Srdjan
+    public $default_language, $current_language;
+
     function __construct () {
 
       self::$instance =& $this;
@@ -41,6 +44,7 @@ if ( ! class_exists( 'TC_plugins_compat' ) ) :
       add_theme_support( 'buddy-press' );
       add_theme_support( 'qtranslate-x' );
       add_theme_support( 'polylang' );
+      add_theme_support( 'wpml' );
       add_theme_support( 'woocommerce' );
       add_theme_support( 'the-events-calendar' );
       add_theme_support( 'optimize-press' );
@@ -86,6 +90,12 @@ if ( ! class_exists( 'TC_plugins_compat' ) ) :
       */
       if ( current_theme_supports( 'polylang' ) && $this -> tc_is_plugin_active('polylang/polylang.php') )
         $this -> tc_set_polylang_compat();
+ 
+      /*
+      * WPML
+      */
+      if ( current_theme_supports( 'wpml' ) && $this -> tc_is_plugin_active('sitepress-multilingual-cms/sitepress.php') )
+        $this -> tc_set_wpml_compat();
 
       /* The Events Calendar */
       if ( current_theme_supports( 'the-events-calendar' ) && $this -> tc_is_plugin_active('the-events-calendar/the-events-calendar.php') )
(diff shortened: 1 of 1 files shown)


Nikeo:
OK let's wait for their reply then.

In the mean time, can you please let me know the general idea :
- how WPML works ?
- how you've been solving the issues both for Customizr and the pro addons ?

I've been looking at your code but I don't get the general principles behind the compatibility patch :)
Thanks !

eri-trabiccolo:
:D 
The main principle is that we cannot use their way to directly filter the theme options using the xml.
This because when we retrieve the theme options (user logged in -> https://github.com/presscustomizr/customizr/blob/master/inc/class-fire-utils.php#L231) we save them again (remember the qtranslateX issue?). This is a problem when translating the string based options, because whit their standard method we overwrite the option in the main lang, so we lose it every time just switching in front from a lang to another one.
So the main part is an alternative to their way, we filter the tc_opt_{option} for the "translatable strings"
(addon like fc will just add its strings to translate to the list of the translatable strings).
To translate string options, with our method or with their one, you need their addon wpm-string-translation. With their method you'll see "ugly" Domain and Name for those strings to translate like:
`admin_text_tc_theme_option[option_name]`, something like that. So since I was there I decided to give to those columns (which are also used as "key" for their translation cache) a better name:
Domain: 'Customizr-option', Name: {Understandable name for the option}.
So these two values must be reachable both from admin (when register the string) and from front (when retrieving the string).
[ Look that for customizr pro 'Customizr-option' become 'Customizr-Pro-option', this means that users migrating from Customizr to Customizr-Pro have to set the translations again, if we don't want this, we should stay with 'Customizr-option' in both cases ]
Hope the "string" part is understandable.

The other part is about featured pages, cats pickers, slider of posts etc etc.
[the bundle isn't wpml ready yet, but it just will use filters hooked to the core, so will be ready as soon as we fix the slider "enhancements" issues]
Simply:
With WPML if you enter the Customizer with "Show all languages" you'll still see some options (like the dropdown-pages you can see when chosing the static front page, but also the categories dropdown we use as base for the cat picker) filtered for the Default language. When entering it with non default language (you can change the language with the lang dropdown select in the admin bar before entering the customizer, from the dashboard) these are filtered for the current language.
So while the available options in the dropdowns were automatically filtered for, say, the current language, the actual selected option wasn't, clearly. Hence, this callback on `option_tc_theme_options`:
`function tc_wpml_customizer_options_transpose( $options ) {`
does that, filiters the selected option in the current language.
This means that, but this is valid also in any case, it's not a special case I introduced :D, that page A (en) will become page A (it) [if A(it] is a translation of A(en)] and in that case we save that translated value, 'cause there's a correspondence, there must be, between the two page IDs (wpml keeps track of it) so we can have these kind of options which are not string based, saved in whatever language in our options, it doesn't really matter. This is useful just if you use to enter the Customizer with different languages, just to not see "Non selected" options.
Anyway all these options are "translated" in front using tc_opt_{option} hook filter.
Other things ?
Posts and Cat filterings (slider of posts, cat restriction for the blog's posts) are handled basically the same way we handle them with polylang:
1) cat: filter the array of categories and translate the cat ids to the current language
2) slider of posts: do not use the transient + alter the slider of posts query with a custom JOIN to look just for posts in the current language.

FPU does a similar thing (less convoluted), just for string based options and for the featured pages ids (in that case, they're just not pages, they can be also other post types).

Anyway there are also other 'caveats' commented in the code.

Hope this is understandable.
So:
1) it's not a real patch of their code, it's a workaround of their way to translate what they call "admin texts"(or strings, don't remember :D). So you cannot see this 'string translation' part in their provided compatibility class
2) It totally keeps their slider compat code, but I want to look further more into this and propose you a "decision" :D 
3) The only "patch" to their code is about the fp ids, which they forced to be pages - if I haven't missed anything - for FPU (in customizr-pro), while we know they can be also other kind of post types.
4) It adds the whole Customizer options translations for not string based options (so page/post IDs for fp and fpu, and cat IDs for pickers)
5) It adds the compat code for custom query we use for slider of posts

Let me know if I need to be more clear!
And.. good morning ;)

Nikeo:
OK. Thanks :)

eri-trabiccolo:
So we keep the slider thing as is for now?

Nikeo:
yes. If as long as you can confirm that there's no risk of regression if we push it like this for now, we'll push a patch for the slider later if needed

eri-trabiccolo:
Mmm, I meant the different way to treat the sliders translation.
But you're probably talking about this ? -> https://github.com/presscustomizr/customizr/pull/403
?

Nikeo:
No I need to be sure that we can push this first set of WPML compat and maybe push a patch later for the slider WPML compat

eri-trabiccolo:
So I'm not sure .. what I have in mind shouldn't break things
but I think that:
a) if I see it will I'll not do it
b) if we still want to do that, we'll inform users, who shouldn't be that many.

Nikeo:
OK fine.
I'll push the new release with your current PR.

eri-trabiccolo:
What about the slider enhancement PR?

Nikeo:
I want to take my time to test it. Even without the meta box revamp.

eri-trabiccolo:
Ok Nico, hopefully we'll not have problems.

Nikeo:
yep :)

PR #388, opened 2015-11-08
Slider enhancements 1

eri-trabiccolo:
re-push #370 without the meta-boxes part
will push the meta-boxes part in another PR so probably we can better debug it.


Nikeo:
Yes good to separate those two.

merged into the slider-enhancements branch

PR #411, opened 2015-11-20
Fix: show second-menu dropdown-submenu arrow anyways

### LINKED ISSUE #410: Secondary menu dropdown-submenu not showing the arrow next to it, when ww < 980px
This is because of this rule:
https://github.com/presscustomizr/customizr/blob/master/inc/assets/less/tc_custom_responsive.less#L437

We should keep it for our secondary menu.

eri-trabiccolo:
fixes #410

### DIFF
--- a/inc/parts/class-header-menu.php
+++ b/inc/parts/class-header-menu.php
@@ -628,6 +628,9 @@ if ( ! class_exists( 'TC_menu' ) ) :
             -moz-border-radius: 6px;
             border-radius: 6px;
           }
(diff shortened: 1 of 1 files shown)


Nikeo:
so concise :)
Thanks !

eri-trabiccolo:
Ops,
this is not enough when submenus revealed on hover because of this:
https://github.com/presscustomizr/customizr/blob/master/inc/assets/less/tc_custom_responsive.less#L226
I think we should act on the less file, what do you think?

eri-trabiccolo:
see #422

PR #413, opened 2015-11-20
Fix: fix some issues regarding to smartload and centering imgs

### LINKED ISSUE #412: Front JS - some issues regarding smartload and centering
1) When smartload enabled the simple load event was triggered on the images cause `_.filter` doesn't return a proper jQuery set but an array, this made the following:
https://github.com/presscustomizr/customizr/blob/master/inc/assets/js/parts/_main_base.part.js#L253
behave incorrectly, as the _ind (index) in that case result the image itself and then _img (image element) was null.
2) Because of a BIG typo whether or not smartload enabled, with this -> https://github.com/presscustomizr/customizr/blob/a5417cac09f9326d4432db563174a88405eeddd0/inc/assets/js/parts/_main_jquery_plugins.part.js#L39
we triggered the simple load ONLY on the holders.

Both issues introduced by ME here:
 https://github.com/presscustomizr/customizr/commit/a5417cac09f9326d4432db563174a88405eeddd0#diff-5d9865b40ef47951c1b36b4bfb0e361a
the error are

eri-trabiccolo:
fixes #412

p.s.
I added some comments to better understand in the future what and why we did.

### DIFF
--- a/inc/assets/js/parts/_main_jquery_plugins.part.js
+++ b/inc/assets/js/parts/_main_jquery_plugins.part.js
@@ -18,28 +18,38 @@ var czrapp = czrapp || {};
     imgSmartLoad : function() {
       var smartLoadEnabled = 1 == TCParams.imgSmartLoadEnabled,
           //Default selectors for where are : $( '.article-container, .__before_main_wrapper, .widget-front' ).find('img');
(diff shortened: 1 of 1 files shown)


Nikeo:
Great thanks for commenting this. Really a good practice.

PR #417, opened 2015-11-24
Fix: remove unneeded css code for the secondary menu dropdown top arrow

eri-trabiccolo:
This is unneeded 'cause we just need to make the :before pseudo element visible (we do it), the other rules will be taken from the main css.
Also this caused an issue in RTL websites as the actual property to set at 9px will be "right" and not "left"

### DIFF
--- a/inc/parts/class-header-menu.php
+++ b/inc/parts/class-header-menu.php
@@ -567,14 +567,6 @@ if ( ! class_exists( 'TC_menu' ) ) :
             -moz-background-clip: padding;
             background-clip: padding-box;
           }
-          .tc-second-menu-on .navbar .nav>li>.dropdown-menu:before {
-            border-left: 7px solid transparent;
-            border-right: 7px solid transparent;
-            border-bottom: 7px solid #ccc;
-            border-bottom-color: rgba(0,0,0,.2);
-            top: -7px;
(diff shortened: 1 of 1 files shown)


Nikeo:
OK did not test it.
Merge it "blindly" :)

eri-trabiccolo:
I tested it :)

eri-trabiccolo:
mm there are some problems here with the latest merge.
Gimme one minute

Nikeo:
I did not build it
Let me create a build a re-push it

Nikeo:
done

eri-trabiccolo:
Not about the build as they're php files.
Something missing, submenus are weird. I'm on it.

Nikeo:
ah yes this is a php inline style stupid me.
Ok wating for your feedback

eri-trabiccolo:
Ah ok, more css is needed for the current active submenu (we need some padding overall, didn't notice 'cause it's fine when the submenu is not the active page - anyway not introduced by this PR ;)).

Nikeo:
ok.

eri-trabiccolo:
> I tested it :)
last famous words :)
Btw I did test it, but I forgot to test when opening on click

see #422

PR #420, opened 2015-11-26
Add: woocommerce cart in the header

### LINKED ISSUE #409: WooCommerce compatibility : header cart option
Since WooCommerce is becoming a major part of WordPress and will probably be integrated into the CMS shortly, the Customizr theme must progressively improve its compatibility with it.

# General Idea #
## Front End ##
The first compat improvement (requested several times by our users) wil be to add a header cart option, just like in the storefront theme.
cart

If WooCommerce is enabled, a shopping cart will be displayed on the right of the tagline. => important, the cart should not be "tagline dependant". If the user decide to hide the tagline, this won't have any impact on the cart. The only way to hide the cart will be to disable it from the WooCommerce new customizer panel (see below).

## Back End ##
If WooCommerce is enabled, a new panel of option named "WooCommerce" will be displayed in the customizer. It will include one section name "Header Cart". This section will include only one checkbox option named "Display the shopping cart in the header"

# Code #
The following code example is the way the StoreFront theme handles the cart in storefront/inc/woocommerce/template-tags.php.

No need to re-invent the wheel there, this kind of code is enough.
The code will be added into the plugincompat class, under WooCommerce.

```php
/**
 * Display Header Cart
 * @since  1.0.0
 * @uses  is_woocommerce_activated() check if WooCommerce is activated
 * @return void
 */
if ( ! function_exists( 'storefront_header_cart' ) ) {
	function storefront_header_cart() {
		if ( is_woocommerce_activated() ) {
			if ( is_cart() ) {
				$class = 'current-menu-item';
			} else {
				$class = '';
			}
		?>
		<ul class="site-header-cart menu">
			<li class="<?php echo esc_attr( $class ); ?>">
				<?php storefront_cart_link(); ?>
			</li>
			<li>
				<?php the_widget( 'WC_Widget_Cart', 'title=' ); ?>
			</li>
		</ul>
		<?php
		}
	}
}
```

# Style #
I did not really investigate the way storefront styles the cart. Some styling might be included in WooCommerce plugin.
The idea is to stick as much as possible with the existing Customizr theme style, and to keep the additional css code as concise as possible.

eri-trabiccolo:
fixes #409

### DIFF
--- a/inc/class-fire-plugins_compat.php
+++ b/inc/class-fire-plugins_compat.php
@@ -912,6 +912,175 @@ if ( ! class_exists( 'TC_plugins_compat' ) ) :
       function tc_woocommerce_change_meta_boxes_priority($priority , $screen) {
          return ( 'product' == $screen ) ? 'default' : $priority ;
       }
+
+      //woocommerce options in the Customizer
+      //add new panel
+      add_filter( 'tc_add_panel_map', 'tc_woocommerce_popul_panel_map' );
+      function tc_woocommerce_popul_panel_map( $panel_map ) {
+        $tc_woocommerce_panel_map = array(
+            'tc-woocommerce-panel' => array(
+                  'priority'       => 50,
+                  'capability'     => 'edit_theme_options',
+                  'title'          => __( 'Woocommerce options' , 'customizr' ),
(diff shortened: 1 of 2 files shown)


Nikeo:
bloody clean code.

eri-trabiccolo:
Do you think so? :D Great

Issue #427, opened 2015-11-28
Turkish language file named wrongly


denzelchia:
Hi,

With reference to this thread on WordPress.org forum.
https://wordpress.org/support/topic/turkish-language-7?replies=7#post-7716296

The language file tr.mo and tr.po should be named as tr_TR.mo and tr_TR.po

Thanks!

Nikeo:
thanks for reporting Denzel :)

eri-trabiccolo:
Thanks for this Denzel.
This is a tricky topic, how wordpress handles translations?
I read somewhere someone complained about the fact that wordpress doesn't fall back on a "similar" lang in those kind of situation. For example, if an user has just `tr` as lang code those translations will be picked up, if he has `tr_TR` they will be not, but also if we rename them to `tr_TR` we'll have the same problem (reversed).
I tend to agree with those complains, Wordpress should take care about a sort of fall back.
Also, sure Nicolas hasn't translated the theme in Turkish, hence someone did it, hence this guy, supposedly tested it, this means that he/she has `tr` and not `tr_TR` as lang code.
Looking here though -> https://make.wordpress.org/polyglots/teams/
the correct lang code for Turkish is actually `tr_TR` but what if there are users having just `tr` as lang code? We'll have complains from these like "After latest update my website isn't localized anymore" :)
So, IMHO, we should keep the `tr` files and add also `tr_TR` (just a copy of `tr`).
What do you think guys?

Nikeo:
OK.
Except the fact that it will add some kbytes to the theme, I think it's the wiser solution and have nothing against it.

eri-trabiccolo:
Mmm something like 100K ?, could this be a problem?

Nikeo:
no

eri-trabiccolo:
:D

eri-trabiccolo:
My poedit tells me that there's also a tr_CY ( Turkish - Cyprus ) that's the problem when you can't fall back !!!

denzelchia:
Hi Rocco,

Login to WordPress admin -> Settings -> General -> Site Language
Use developer tools to look at the option values of that drop-down list.
Only those on this list is supported by WordPress, tr_CY is not supported.

When user select the Site Language at this option, WordPress will find the relevant language file within the Theme and load it. If WordPress cannot find the language file with name similar to the selected option value, it will not load any Theme translation.

Hope this helps. :)

Thanks!
Denzel

eri-trabiccolo:
[...]

PR #429, opened 2015-11-28
Fix: remove unsupported turkish translation files

eri-trabiccolo:
see #427
must be merged with #428

according to:
https://make.wordpress.org/polyglots/teams/
https://developer.wordpress.org/reference/functions/wp_get_available_translations/
the correct Turkish lang code is `tr_TR` (added in #428)
lang code `tr` is not supported by wordpress.
As Denzel suggested in #427 we should stick with the wp supported languages.
Anyway the boss will decide on this 'cause he, tautologically, is THE BOSS :P


Nikeo:
this translation is not supported anymore, therefore it is not supported anymore
:)

eri-trabiccolo:
ahah Tautology rules!

PR #431, opened 2015-11-29
Fix: fix special case when a submenu dropdown might overlap menu items

eri-trabiccolo:
Say we have a drodown with 2 menu items X and Y.
X has has a dropdown children, Xd.
Xd must be moved 'cause it overflows the window.
We have to be sure that the new offset will not make Xd overlap Y.
Hence we leave like 50px to let the user be able to pass from X to Y


eri-trabiccolo:
Don't merge this yet. Jusy thought it's incomplete. Will finish it tomorrow morning. Sorry

eri-trabiccolo:
(should be fine now)

Nikeo:
yep, good point :)

PR #435, opened 2015-11-30
Fix: allow logo transition only when both w,h are available.

eri-trabiccolo:
We should avoid setting them not just when both width and height
are not valid numbers, but also when one them is not a valid number.
This simply means change the exit condition from

0 === _.size(w,h)

to

2 != _.size(w,h)

I never met this issue, but still I think this should be the correct
condition.


Nikeo:
yes makes sense !

PR #437, opened 2015-11-30
Fix: sticky-header push div always rendered in live preview

### LINKED ISSUE #436: Customizer Preview: Sticky header not correctly rendered
That's the issue:
Be sure the sticky header option is not enabled in the Customizer
- Go in Appearance -> Customize (or refresh it if you had to disable the sticky header option)
- Enable the sticky header.
You'll see the "main-wrapper" slipping behind the header.
That's because we don't print the push div (#tc-reset-margin-top) if the option is not enabled.
This regression might have been introduced in the header/menu overhaul.

eri-trabiccolo:
It'll be displayed according to the option status.
fixes #436


Nikeo:
ok. got it

PR #438, opened 2015-12-01
Fix: better rendering of the slider loading gif

eri-trabiccolo:
This includes:
1) js
A fadingOut when disposing the loading div. Also, since we can have
more than one slider per page, there's no reason to wait for all the
slides of all the sliders have been parsed by the centerImages plugin
so we dispose the loading div per slider.
2) php/css
The loading image is now rendered as background. This allow us to get
rid of some CSS code but keeping the possibility for a user to specifiy
a different loading icon using the same filter.
Alternatively we might move this small CSS code to the main CSS file
and if users want to use a different gif they have to override it.


Nikeo:
works pretty good. Makes me want to enable it by default :)

eri-trabiccolo:
Hehe the fade out really makes the difference

PR #441, opened 2015-12-02
Fix: rtl colophon fixes

### LINKED ISSUE #440: RTL colophon issues
1) RTL colophon blocks are not in the right order.
RTL Colophon should look like the mirrored version of the LTR one.
So socials on the right and back to top on the left (as well as the btt button), atm it looks like this:

fig a)
rtl-colophon

This is because when RTL we reverse the blocks order with php but we also set the bootstrap floating direction to right => annihilation
Another side effect is that if we remove the btt column (which is what we want to do when the btt button is enabled) you'll see the following:

fig b)
rtl-colophon-no-btt

This is also because the social-block still keeps its `pull-left` class, so we'll have
span3(pull-left) - span3(nothing) - span6(float: right), while it should be
span3(nothing) - span6 - span3(pull-right).

2) Credits elements are messy (see fig a). "Powered by W" should be on the left and "Customizr" on the right. Also "Designed by Press Customizr" as it's composed by two "blocks" ("Designed by" and "Press Customizr") should look like "Press Customizr Designed by"

eri-trabiccolo:
1) php : as in RTL we change the bootstrap column floating from left to
right, there's no need of the blocks reordering. Though we change
some blocks classes according to direction
(pull-left to pull-right, and vice versa)
2) css: in RTL btt button should be shown on the left (mirroring)
3) php/css: fix the messy RTL credits performed by:
- wrap credit texts in span tags (php)
- display span tags above and links as "inline-blocks"
fixes #440

### DIFF
--- a/inc/assets/less/tc_custom.less
+++ b/inc/assets/less/tc_custom.less
@@ -3043,7 +3043,10 @@ footer#footer .colophon {
 footer#footer .colophon a, footer#footer .colophon p {
   color:@FooterBg;
 }
-
+footer#footer .colophon .credits span[class*=text],
+footer#footer .colophon .credits a{
+  display: inline-block;
+}
 
 
 footer .widget_recent_entries li:after {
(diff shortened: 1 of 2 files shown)


Nikeo:
ok. Thanks

PR #442, opened 2015-12-02
Fix: do not display colophon's btt text when btt arrow button enabled

### LINKED ISSUE #394: Remove the additional 'Back to Top' Link when Back to top arrow is enabled
This was already talked about in a past issue I think.
Anyway, this was reported here : https://wordpress.org/support/topic/additional-back-to-top-link-after-update?replies=2

eri-trabiccolo:
fixes #394


eri-trabiccolo:
p.s.
the indentation of this class is a mess :) 
Didn't want to touch it (even if something went wrong anyways :)) otherwise understand what I really changed would have been hard.
Anyway I think we should clean the code, maybe with the new framework, from this standpoint :)

Issue #443, opened 2015-12-02
Customizer Controls: Alternate Thumbnail's height control visibility


eri-trabiccolo:
This control should be visible only when:
a) Display the post thumbnails ON
AND
b) Thumbnails shape is rectangular-*

If we enter the Customizer when b) but not a) we'll still see the "Alternate Thumbnail's height control", due to the fact that condition b) is processed after condition a).
Notice: when processing condition a) we set the visibility of the control in object depending on the condition b) too (cross-dependency).
The solution would be, as we already did for similar cases, to swap the order of a) and b) in _control.js .

PR #444, opened 2015-12-02
Fix: alternate layout is available only if thumb position is right/left

### LINKED ISSUE #362: Disable Alternate thumbnail/content when thumbnail position is top or bottom
Option located in the customizer > Content > Post list > Post list design.
If thumbnail position == top || bottom, it doesn't really make sense to keep the alternate design on.

* customizer :  hide the alternate thumbnail/content control in the customizer
* front end : filter alternate to false.

eri-trabiccolo:
Also fixes the alternate layout thumbnail's height control visibility,
see(fixes) #443

fixes #362

### DIFF
--- a/inc/admin/js/parts/_control.js
+++ b/inc/admin/js/parts/_control.js
@@ -285,6 +285,22 @@
         }
       }
     },
+    'tc_post_list_thumb_shape' : {
+      controls: [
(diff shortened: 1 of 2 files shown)


Nikeo:
great, thx for the additional fix

PR #448, opened 2015-12-03
Fix: allow smartload in wp 4.4 (responsive images) + small fix

eri-trabiccolo:
1) Basically we replace the attribute srcset with data-srcset in php
when parsing the imgs. In front js we'll swap them again (as we already
do with the src attribute)
2) Unrelated small fix: I noticed that the load function sometimes is
called twice (for the same image), just added an exit condition when the
image has already been smartloaded (has class tc-smart-laoded)


Nikeo:
coool

PR #461, opened 2015-12-15
Fix: apply border bottom only to theme sidebars widget list item

### LINKED ISSUE #459: Border bottom in widget lists
We should limit the CSS rule which applies a border bottom to the widget list items in the #main-wrapper just to the sidebars.
This cause there are several plugins which use the widget system putting widgets inside the content (see page builder & co).
What do you think?

https://secure.helpscout.net/conversation/149415956/5002/?folderId=607546

eri-trabiccolo:
to avoid styling widgets in the #main-content which are not in
sidebars (for example those printed by some plugin in the content)

fixes #459

### DIFF
--- a/inc/assets/less/tc_custom.less
+++ b/inc/assets/less/tc_custom.less
@@ -1478,13 +1478,13 @@ font-size: 13px;
   }
 }
 
-#main-wrapper .widget li {
+.tc-sidebar .widget li {
   line-height:25px;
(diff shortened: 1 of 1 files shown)


eri-trabiccolo:
(Travis error isn't due to my code :D)

Nikeo:
useful thanks

PR #462, opened 2015-12-15
Fix: fix smarload preg callback - reverse strpos param order

eri-trabiccolo:


Nikeo:
wow, strange we just find this now. I'm surprised this had no other side effects!

eri-trabiccolo:
Mmm, well, the thing is that I recently introduced it (latest version) + it should not have big side effect just potentially parsing the imgs twice (maybe an issue with the holders too).
:|

Nikeo:
not a big deal. Thanks for the fix :)

PR #471, opened 2015-12-17
Fix: better html comments

### LINKED ISSUE #470: Closing main-wrapper div html comment, can be seen as a server directive
Issue raised here:
https://secure.helpscout.net/conversation/150627404/5201/?folderId=607709
error message:
[an error occurred while processing this directive]

I'm unable to reproduce it but I can see why it might happen:
http://httpd.apache.org/docs/1.3/mod/mod_include.html

I see in the user's page giving that error that:
    </div>

gives no issues. So we can safely use that form.

eri-trabiccolo:
fix rare cases when some html comments were considered as server's
directives.
fixes #470

### DIFF
--- a/comments.php
+++ b/comments.php
@@ -27,4 +27,4 @@ if ( have_comments() ) {
 		if ( have_comments() && apply_filters( 'tc_display_comment_list', true ) )
 			do_action ( '__comment' );
 	?>
-</div><!-- #comments .comments-area -->
+</div><!-- //#comments .comments-area -->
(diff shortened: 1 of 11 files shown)


eri-trabiccolo:
Hope I haven't forgot any other similar comment
(as you can see, to be sure I was a bit redundant)

Nikeo:
good fix thanks. I'll keep that in my alzheimer mind

eri-trabiccolo:
Ahaha I think you already knew that as in some places you put a safe comment. :D

PR #473, opened 2015-12-21
Fix: better handling of the theme (default and custom) options

### LINKED ISSUE #241: Slider demo height
Reported here:
https://wordpress.org/support/topic/top-of-slider-cut-off-1?replies=6#post-7333605

I just installed a brand new customizr and the demo's height is 750px.
The thing is that as soon as you ENTER the customize the default 500px is stored in the db (as well as all the other options) and then this value is applied. (am I wrong?)
Why we don't store just the delta between default and new options as we did before?
Can this depend on these improvements (?):
https://github.com/Nikeo/customizr/commit/0a2778c33b4ebafa82951bb9c6dff49fc99b538c
https://github.com/Nikeo/customizr/commit/d720a8401a35631eb2c11cc3c4bbc3c67584b360

Mmm maybe not.
Looks like this happens as soon as you activate the theme.
Try this:
1) switch on another theme
2) delete tc_theme_options from the wp_options table in the db
3) activate customizr from Appearance -> themes
4) look at the tc_theme_options field in the db
it has all the options! (not as defaults array)

Now do
2)
3a) go on a site page
4) look at the tc_theme_options field in the db:
it has the defaults array

eri-trabiccolo:
should fix #241

### DIFF
--- a/inc/class-fire-utils.php
+++ b/inc/class-fire-utils.php
@@ -1053,11 +1053,12 @@ if ( ! class_exists( 'TC_utils' ) ) :
     * @param $_defaults (optional), the array of options to remove from the $_options
     * @param $_store    (optional) default false , whether or not store the $_options in the db
     * @param $_add_defatuls (optional) default false, whether or not add the array of default options as element of the array of options 
+    * @param $_force (optional) default false, whether or not force the sanitation 
     *
     * @return array
     * @since v3.4.18
     */
-    function tc_maybe_sanitize_theme_options( $_options = null, $_defaults = null, $_store = false, $_add_defaults = false ) {
+    function tc_maybe_sanitize_theme_options( $_options = null, $_defaults = null, $_store = false, $_add_defaults = false, $_force = false ) {
       //In versions < 3.4.18 (since "unknown" atm :) ) we merged also defaults options
       //hence we have to get rid of them
       //Conditions are:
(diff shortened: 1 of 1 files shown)


eri-trabiccolo:
Ah, no I made an error, actually the diff checks the values. So this PR is wrong!!!
I'll look at it tomorrow morning

eri-trabiccolo:
So yeah, we need to remove from the "merged_options" the defaults one by one, potentially recursively if there are default options which are array not empty. But honestly .. cannot see not empty arrays which are not empty, anyways we better expect this as we don't know whether or not will add them, and if an user will upgrade from a version < 3.4.17 to the version when we'll use them.

Also consider that we do this "removal" in tc_set_option only which is called very rarely .. but for sure is called when you upgrade the theme.

About the way to check whether or not we have to do the removal. It's not very good atm, I agree, 'cause, even if it's rare, there could be the case when we introduce a lot of new defaults options so that this will fail, so what we need.. I think, is to put a "marker" in the options once the removal is done. What do you think?

Tomorrow I will fix the wrong diff.

eri-trabiccolo:
So the thing is that I think I pushed the wrong PR (a staging branch). I actually used array_diff_assoc in the final one. Which anyways isn't good either since when comparing the values it casts them to strings. So we better cycle among the defaults and check with a simple "==" if their equal to those in the "wrong" options, if so we unset them in the "wrong" options.
Honestly I would like to add a marker like "options_sanitized" as new key in the theme_options, which will allow us to better check whether or not we have to sanitize them.
Will push a PR later today which will overwrite this one.

eri-trabiccolo:
PR redone.
What do you think?
Let me know

p.s.
sorry for the late fix I see you already released a new version

eri-trabiccolo:
I added a further parameter.
This can be useful in the following edge case:
- Update to the new version => sanitation occurs => options_sanitized option added
- Revert to a previous version => options messed up again
- Update to the new version again => sanitation doesn't occur because of the presence of "options_sanitized"

I think we could add a link in the update notice to perform this action "manually" as a lot of plugins do to upgrade the db...

What do you think?

eri-trabiccolo:
Please take another look at it

eri-trabiccolo:
Fixed in:
https://github.com/presscustomizr/customizr/commit/86fc1072cfd6f0123e7d67b5d4ef20fd66d0fef9

PR #476, opened 2015-12-24
Fix: fix disabling wc-header-cart to reset tc_user_options_style

eri-trabiccolo:


eri-trabiccolo:
of course.. my fault, sorry.
reported here: https://wordpress.org/support/topic/woocommerce-cart-header-causing-problems
(cannot reproduce a different plugin compatibility issue reported there, atm)

Nikeo:
ok thanks

PR #478, opened 2015-12-28
Fix: fix btt-arrow and scroll-down issue

### LINKED ISSUE #477: Btt-arrow: when scrolling to the top, for a small amount of type you cannot scroll-down
You can easily reproduce it on the demo site.
1) Let the btt-arrow appear (scrolling down the page for 100px + )
2) click on the btt-arrow
3) wait till the page reaches the very top
4) rapidly start to scroll-down the page 

You can see "a force" which opposes to that :) Very ugly.
This doesn't happen with the btt link in the colophon.
From my understanding (and tests) the culprit line is:
https://github.com/presscustomizr/customizr/blob/master/inc/assets/js/parts/_main_userxp.part.js#L114

Fom my undertanding and tests ( :) ) as the `.btt-arrow` is inside the `.tc-btt-wrapper` the click event propagates we get two "out-of-phase" reactions which causes the issue.
The solution might be :
a)limit the element's classes to just `.back-to-top` and `.btt-arrow`
or
b)use the jQuery function event.stopPropagation();
inside the click/touch callback

I prefer the latter.

eri-trabiccolo:
Also use more descriptive variable names.
fixes #477

### DIFF
--- a/inc/assets/js/parts/_main_userxp.part.js
+++ b/inc/assets/js/parts/_main_userxp.part.js
@@ -107,11 +107,13 @@ var czrapp = czrapp || {};
     //BACK TO TOP
     backToTop : function() {
       var $_html = $("html, body"),
-          _backToTop = function($) {
-            return ($.which > 0 || "mousedown" === $.type || "mousewheel" === $.type) && $_html.stop().off( "scroll mousedown DOMMouseScroll mousewheel keyup", _backToTop );
+          _backToTop = function( evt ) {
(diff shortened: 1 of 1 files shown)


MikeHoltRA:
I copied the code into my theme at \customizr\inc\assets\js\parts tu I still have the same problem...the scroll to top button is not displaying. Can anyone help me resolve this please? You can see the problem here: http://advance-australia.com.au/

eri-trabiccolo:
Hi @MikeHoltRA 
this fix is already included in the version you're running on your website 3.5.12
Also copying the code in that file won't help as it's a dev file which will be deployed in tc-scripts.js and tc-scripts.min.js

Would you mind opening a new ticket on the [theme support forum](https://wordpress.org/support/theme/customizr#new-post) explaining what's the real issue you have?

Thanks

PR #480, opened 2015-12-29
Fix: customizer: don't remove wp background customization controls

### LINKED ISSUE #479: Custom Background: why do we remove the customizer settings?
Why?
Isn't not declaring `add_theme_support('custom-background')` enough?
With wp4.4 (at least) looks like it's enough, so we could even remove these lines:
https://github.com/eri-trabiccolo/customizr/blob/master/inc/class-fire-utils_settings_map.php#L2187
https://github.com/eri-trabiccolo/customizr/blob/master/inc/class-fire-utils_settings_map.php#L2189

Am I missing anything? (I should check old wp versions..)

https://secure.helpscout.net/conversation/154525158/5718/?folderId=607709

eri-trabiccolo:
They will be shown only if users adds
add_theme_support('custom-background')
to the child-theme functions.php

fixes #479

### DIFF
--- a/inc/class-fire-utils_settings_map.php
+++ b/inc/class-fire-utils_settings_map.php
@@ -2183,9 +2183,7 @@ if ( ! class_exists( 'TC_utils_settings_map' ) ) :
     function tc_popul_remove_section_map( $_sections ) {
       //customizer option array
       $remove_section = array(
-        'background_image' ,
         'static_front_page' ,
-        'colors',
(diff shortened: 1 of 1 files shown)


Nikeo:
ok.
anyway that's something I'd like to implement in the future as already mentioned. Probably in Pro.

eri-trabiccolo:
Yes I remember :)

PR #481, opened 2015-12-30
Imp: some slider loader improvements

eri-trabiccolo:
1) replace gif loader with a pure CSS loader
still use the gif as fall-back for old browsers (those which don't have
csstransforms3d). Several php filter have been added so we can easily
revert the pure CSS loader to the standard gif if users do not want it
( it can be a pro feature, we might also consider to let user load their
own gif loader ). Minor side effect: as the fall-back is for those browsers
with no-csstransforms3d (class added to the html by modernizr), the gif
with a small delay as it depends on the modernizr processing. Not a big
deal anyway, I think we shouldn't care too much about that.

2) add a little small delay before fading out the loader (500ms),
because of the re-sizing which might occurr after it has been disposed
when dealing with retina imgs (retina.js)

### DIFF
--- a/inc/assets/js/parts/_main_jquery_plugins.part.js
+++ b/inc/assets/js/parts/_main_jquery_plugins.part.js
@@ -129,8 +129,13 @@ var czrapp = czrapp || {};
             defaultCSSVal : { width : '100%' , height : 'auto' },
             useImgAttr : true
           });
-          //fade out the loading icon per slider
-          $( this ).prevAll('.tc-slider-loader-wrapper').fadeOut();
+          //fade out the loading icon per slider with a little delay
+          //mostly for retina devices (the retina image will be downloaded afterwards
+          //and this may cause the re-centering of the image)
+          var self = this;
+          setTimeout( function() {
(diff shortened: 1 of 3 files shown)


eri-trabiccolo:
Tested with:
- OSx:
(*) safari 6.1.5
- Win:
(*) chrome 47.0.2526.106 m
(*) ie 11
(*) ie 10/9/8 emulated with ie11 inspect element - ie9 and below don't have csstransforms3d, so they fall back on the standard gif
- Linux:
(*) chrome 47.0.2526.106 (64b)
(*) firefox 42.0
(*) opera  - doesn't have csstransforms3d, so it falls back on the standard gif

Nikeo:
looks overall very clean. Let's see the result!

eri-trabiccolo:
Let me know if you like it :) 
I didn't really change the CSS except for the circle "centering" (the original used the flexbox system)

eri-trabiccolo:
So what do you think about the result?
You didn't put this "enhancement" in the release note on presscustomizr.com :(
ehehe
:P

PR #496, opened 2016-01-06
Fix: remove useless _blank target for some customizer links

eri-trabiccolo:
There's no reason to have the _blank target for customizer links which
are handled via js (pointing to a customizer itself element).
Also the _blank target seems to cause an issue with some browsers:
- with chrome those links worked fine
- with firefox a new empty tab is opened and nothing changes in the main
  one


giorgioriccardi:
Tested on WP4.4 without the _blank and the problem was gone!
Oddly enough Chrome seems to be the only browser able to understand that the _blank target is useless and therefore ignores it.

eri-trabiccolo:
Yeah, weird Chrome is the only one :D 
Thanks for the feedback Giorgio, and of course for raising the issue in the first place :D 
Rocco

Nikeo:
yep good catch thanks guys :)

PR #500, opened 2016-01-10
Fix: fix jetpack's photon - theme smartload compatibility issue

### LINKED ISSUE #493: FW0018 (Issue) Feature "Load images on scroll" doesn't work in blogs
https://wordpress.org/support/topic/feature-load-images-on-scroll-doesnt-work-in-blogs

eri-trabiccolo:
This issue is due to the fact that jetpack's photon, when parsing imgs
server side removes their width/height attributes. It recalculates/sets
them client side (js) when the actual image (due to smartload) is a dummy one
1x1px.

fixes #493

### DIFF
--- a/inc/assets/js/parts/jqueryimgSmartLoad.js
+++ b/inc/assets/js/parts/jqueryimgSmartLoad.js
@@ -138,6 +138,22 @@
       //prevent calling this twice on an already smartloaded img  
       if ( $_img.hasClass('tc-smart-loaded') ) return;
       $_img.fadeIn(self.options.fadeIn_options).addClass('tc-smart-loaded').trigger('smartload');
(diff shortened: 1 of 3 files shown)


eri-trabiccolo:
When do you plan to release a new version?
Jetpack has just been updated - photon implied.
I would like to see if we need something more to fix the compatibility with photon, but I cannot do this right now as I'm about to leave in less than an hour.
Anyway if you plan to release it now it will take some days before wp will put it live so we should have the time to make some other fixes if needed.

Nikeo:
No worries, it will be released early next week.

eri-trabiccolo:
I tested the latest jetpack with 3.4.17
Looks fine.

PR #523, opened 2016-01-25
Fix: fix featured pages text trim

### LINKED ISSUE #320: Featured pages text trim
reported here -> https://wordpress.org/support/topic/customizr-and-qtranslate-in-featured-pages?replies=6
Should be the same fixed in fpu, which I identified as:

> If you have a text >=200 chars with no spaces after the limit strpos returned false, and substr from 0 to false, returned nothing, then you had just "..." displayed.

Solution:
port fpu text-trim code for fp free too (similar to the text trim for the slider of posts, I propose to move the function in utils ).

eri-trabiccolo:
fixes #320


eri-trabiccolo:
How to reproduce the issue and test the PR.
Reproduce the issue:
1)Put the the following text of 202 chars ( without any space after the 200th char ) in one of the featured texts:
>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer in nisi pulvinar, rutrum ex quis, porta mi. Praesent ultricies dolor ultricies, fermentum felis eget, ultricies mi. Duis eget massa nunc.
See the result:
You'll see only the "..." printed as featured text

Test the PR
Switch on this branch and in the same conditions refresh the home page.
You'll see the correct behavior.

eri-trabiccolo:
@Nikeo 
I think this should be done a little better, but I need your opinion 'cause I'm not sure the result we want to achieve.


Our conditions:
Say the limit is 200 chars.
We use to cut the text at the first blank space after 200chars.

Example:
We have a text of 250 chars.
After the 200 char there are no blank spaces.

This PR will cut the text at 200chars if there are no blank spaces after 200chars (1).
But I'm thinking that maybe we want to preserve the "words". In this case this PR is wrong 'cause it should not cut the text at all (2).

Is this understandable?

Let me know 'cause accordingly to this I must change also the fpu trim which, inconsistently, atm respects the case (2)

Nikeo:
ok, let's not cut the words

eri-trabiccolo:
So method (2).
Ok I'll amend this PR (adding multibyte compatibility see: https://secure.helpscout.net/conversation/164259939/7681/)

eri-trabiccolo:
PR updated + added the multibyte patch.

eri-trabiccolo:
same reason, multibyte compatibilty .. don't you like it?

eri-trabiccolo:
Ok I have to re-do this, messed up after a merging.

PR #524, opened 2016-01-25
Fix: fix grid expanded post edit link not reachable

### LINKED ISSUE #286: Grid expanded post edit link not reachable
The edit link in the grid expanded post must be placed outside the summary div so to not be overlapped by the link to the post, as we already do in GC.

eri-trabiccolo:
fixes #286


eri-trabiccolo:
A Grid Customizer update is suggested (PR already pushed), though isn't mandatory, as this PR doesn't have visible UI effects since even if the edit link is printed twice the 'second' button will overlap the 'first'

Nikeo:
ok thanks :)

PR #527, opened 2016-01-25
Add: new option - display blog posts in random order

eri-trabiccolo:
fixes #491
thanks to @menakas

PR #529, opened 2016-01-27
Add: add new option to hide/display post navigation in home/blog

### LINKED ISSUE #528: Customizer Preview: Some options do not work
Just found that the contextual post metas + contextual and general post navigation hide/show options do not work as expected.

For the general post navigation option the reason is simple.
The code misses the element retrieval:
https://github.com/presscustomizr/customizr/blob/master/inc/admin/js/theme-customizer-preview.js#L444
where do we define that var? Nowhere .. 
+
https://github.com/presscustomizr/customizr/blob/master/inc/admin/js/theme-customizer-preview.js#L464

the options prefix isn't needed anymore.

so easy fix

About the other issues:
This line is not right -> https://github.com/presscustomizr/customizr/blob/master/inc/admin/js/theme-customizer-preview.js#L418

Issues probably introduced during the big preview revamp.

eri-trabiccolo:
also fix issue with some contextualized options not working in the
customizer preview (post metas, post/page navigation)
fixes #528

### DIFF
--- a/inc/admin/js/parts/_control.js
+++ b/inc/admin/js/parts/_control.js
@@ -174,32 +174,25 @@
     'page_for_posts' : {
        controls: [
          'tc_blog_restrict_by_cat',
-         'tc_show_post_navigation_home'
        ],
        callback : function (to) {
          return '0' !== to;  
        },
-      //display dependant if master setting value == value
-      cross: {
-        tc_show_post_navigation_home : { master : 'tc_show_post_navigation' , callback : function (to) { return '1' == to; } },
-      }
     },
     'show_on_front' : {
       controls: [
         'tc_blog_restrict_by_cat',
         'tc_show_post_navigation_home'
       ],
-      callback : function (to) {
+      callback : function (to, targetSetId) {
+        console.log(targetSetId);
         if ( 'posts' == to )
           return true;
-        if ( 'page' == to )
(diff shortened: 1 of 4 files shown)


eri-trabiccolo:
I have to say that the small system I created to recognize the context made this add very simple.
I'm satisfied :P

Nikeo:
This is not what I had in mind, I should have defined it better sorry.

It has to be an option available for the blog on home only. not for the blog on another page.
Therefore the is_home() check is not enough.

in the customizer, this option should be located in content > front page and revealed when the user decides to display the blog posts on home.

the name of the option should also reflect this : 'tc_show_post_navigation_home'

eri-trabiccolo:
Ah don't worry.
Basically this is what I had in mind 'cause I proposed it after the random posts PR, your specification came after.
So I haven't been clear ;)

It's basically the same PR.
It should still depend on the post navigation global option right?
And the blog, when not in front, will still be controlled.
So last question.. this -> https://github.com/presscustomizr/customizr/pull/527 
should be valid only for the blog on home either, right?
'Cause the original code provided by menakas impacted on the blog on another page too.
I think I have to change that too..

eri-trabiccolo:
Ok, done.
Check if you like it when you can.
Let me know about https://github.com/presscustomizr/customizr/pull/527 too
even if the change to apply it only in home displaying the blog is very simple

PR #544, opened 2016-02-04
Fix: avoid outline showing up on links click in ff (v44)

### LINKED ISSUE #538: FW0032 Border around links in FF
https://wordpress.org/support/topic/borders-around-all-links-since-update

I raised this with Rocco some time ago - don't remember the solution sorry!

eri-trabiccolo:
fixes #538


eri-trabiccolo:
the small change in the smartload js, it's just a minor "typo" amending .. not related to the issue above.. but since I was there .. :)

PR #550, opened 2016-02-08
Logo from media BETA

eri-trabiccolo:


inc/class-fire-utils_settings_map.php
```diff
@@ -2723,8 +2733,26 @@ function tc_sanitize_number( $value) {
...
+    /**
+     * adds sanitization callback funtion : media attachemnt
+     * @package Customizr
+     * @since Customizr 3.4.19
+     */
+    function tc_sanitize_media_attachment( $attachment ) {
+      $_maybe_numeric_attachment = $this -> tc_sanitize_number( $attachment );
+      if ( $_maybe_numeric_attachment )
+        return $_maybe_numeric_attachment;
+      else {
```
eri-trabiccolo: I tested this "block" (the 'else' should be removed as it comes after a return statement :D ) switching to the 3.0 tag (git), setting a logo (stored as url there) then switched back to the branch of this PR and everything worked fine.

PR #557, opened 2016-02-13
Fix: unify text trimming (fp/slider) + multibyte compatibility

### LINKED ISSUE #320: Featured pages text trim

eri-trabiccolo:
fixes #320
see discussion in #523

### DIFF
--- a/inc/class-fire-utils.php
+++ b/inc/class-fire-utils.php
@@ -30,6 +30,9 @@ if ( ! class_exists( 'TC_utils' ) ) :
         //init properties
         add_action( 'after_setup_theme'       , array( $this , 'tc_init_properties') );
 
+        //multibyte string compatibility
+        add_action( 'init'                    , array( $this , 'tc_multibyte_string_compatibility') );
+
         //Various WP filters for
         //content
         //thumbnails => parses image if smartload enabled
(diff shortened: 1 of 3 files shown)


eri-trabiccolo:
(let me know your opinion, 'cause the same should be done with fpu)

eri-trabiccolo:
do not merge this. I've found an issue with the slider.
We cannot always preserve the words because in edge cases the caption will exceed the slide height (because the default slide css doesn't provide an actual fixed height, but a min-height ).
Will finish this tomorrow.

eri-trabiccolo:
Ok Nico, so this is something you have to decide.
I think we should fix the max-height of the slider item too (so not only the min-height), 'cause potentially we can have the problem described above just increasing the font size of the title/description of the slides. Also consider that we already fix the max-height when using a custom slider's height -> https://github.com/presscustomizr/customizr/blob/v3.4.18/inc/parts/class-content-slider.php#L1249

PR #559, opened 2016-02-15
Fix: slider - avoid caption increasing slides height

eri-trabiccolo:


eri-trabiccolo:
see https://github.com/presscustomizr/customizr/pull/557#issuecomment-183865359

Issue #560, opened 2016-02-15
Add a link to doc in the customizer


Nikeo:
When clicking on the interrogation point the contextual helped is revealed.
![link_to_doc](https://cloud.githubusercontent.com/assets/3038156/13046794/492cb87c-d3db-11e5-8437-f997dc800ea9.png)

The idea would be to always append a link to the doc here.
Php filter if possible or javascript

eri-trabiccolo:
Do not think is possible with php, I cannot see available filters, at least for the "first info" (I mean the one shown above). Have to see for the others.
Would be great having contextual doc links..

andrewfandrew:
Hi, I came across http://justintadlock.com/archives/2011/06/02/adding-contextual-help-to-plugin-and-theme-admin-pages which might be a way to do it using PHP filter.

andrewfandrew:
Hi, I have found the relevant core file. The code that controls the text displayed is at the bottom:
https://core.trac.wordpress.org/browser/trunk/src/wp-includes/class-wp-customize-panel.php

eri-trabiccolo:
Yeah.
That's for the panels though, cannot really use php filter on the "first info" (as I said above, and I meant the info shown as soon as you enter the Customizer, the one displayed in the picture above
https://core.trac.wordpress.org/browser/trunk/src/wp-admin/customize.php#L140
)
Anyway I'm on it, js for the "first info" and panel/section description for panels and section ;).
Let's see if the boss likes it :P

andrewfandrew:
Hi, sorry you were correct: it is a JS not a PHP filter task.

eri-trabiccolo:
Yeah Andrew,
if you're interested you can try the pr above.
p.s.
to actually see a difference in the "first info" you have to run grunt in order to build the back js.
The panel info is, instead, visible since it's only php, in the info block of the Global Settings panel.

andrewfandrew:
Thanks Rocco, That is a useful pointer to use grunt. I have read your code and it is useful seeing how to implement these solutions.

andrewfandrew:
Hi, I have not used travis but just signed up and linked my github account to travis. Rocco I see your code did not pass the travis unit test? I would like to assist. Or if you could explain a little about how you are using travis? Can you give any tips on tracking down issues raised by travis tests?

eri-trabiccolo:
Hi Andrew,
[...]

PR #561, opened 2016-02-15
Imp: allow contextual doc link in the customizer info

eri-trabiccolo:
About #560 
What do you think?
Of course you can put the whole html code in the description field in utils_settings_map.
But I preferred this way so to keep the filling of the panel/settings map more focused on "simple" parameters.


Nikeo:
I like this, let's add it !

PR #596, opened 2016-03-29
Match Finnish translation fi_FI -> fi

daniellandau:
Wordpress core uses `fi` instead of `fi_FI` so rename the translation file name here to match core. Also I translated some of the missing strings.


Nikeo:
merged in dev.
thanks !

PR #617, opened 2016-05-14
Imp: remove it lang files - will use the ones available on wp.org

eri-trabiccolo:
It lang pack is now available on wp.org and automatically downloaded.
We should not ship it anymore in the theme.


Nikeo:
fixed in dev.
thx

PR #639, opened 2016-09-14
Customizr tec

eri-trabiccolo:


eri-trabiccolo:
Don't think that those intermediate commits are a problem, but let me know otherwise.

Issue #668, opened 2016-10-28
Missing German Language file in Theme ( Customizr Pro Theme )


denzelchia:
Hi,

This is an issue for Customizr Pro Theme.
Customer reported that German Language files are missing in recent updates.
https://secure.helpscout.net/conversation/269544953/18733/?folderId=631714

Customer send us a gentle reminder, to remember to include de_DE.mo and de_DE.po  in the next Customizr Pro Theme update.

Thank you very much for your attention. :)

Best Regards,
Denzel

eri-trabiccolo:
Thanks Denzel
we're working on a fix.

Best regards,
Rocco

denzelchia:
Hi Rocco,

You are welcome!

Thank you. :)

PR #670, opened 2016-10-28
Imp: add back langpacks completed on translate.wordpress.org

eri-trabiccolo:
Needed for Customizr-Pro
note: german GTE considered our german translation as formal
german. The current de_DE.po(mo) is just a copy of the
de_DE_formal.po(mo) pack


eri-trabiccolo:
Fixes #668 
and https://github.com/presscustomizr/customizr-pro/issues/9

PR #681, opened 2016-11-07
Dev c4 final

eri-trabiccolo:


Nikeo:
...this is what I call a short PR...

still feeling like a sidebar. sorry but you're the one of the few who can
understand that.

eri-trabiccolo:
I really would love to knoe how does it feel?  :(

PR #683, opened 2016-11-14
Imp: update FontAwesome to v4.7.0 + add Snapchat social link

### LINKED ISSUE #682: [Enhancement] Snapchat social button
As per this forum request, it would be good to have Snapchat social link in the default set.
[https://wordpress.org/support/topic/snapchat-social-button/#post-8431852](https://wordpress.org/support/topic/snapchat-social-button/#post-8431852)

Considering that snapchat is gaining in popularity, we may want to consider it.

eri-trabiccolo:
fixes #682


eri-trabiccolo:
Test done, no regression

PR #696, opened 2016-12-05
Dev c4

eri-trabiccolo:


Nikeo:
Hi Rocco, for our future PR would it be possible that you limit the number of commit to around 10-15 by PR, so that it's easier to check the changes ?
Thanks !

PR #718, opened 2017-01-03
Imp: encode pipes when requesting multiple gfont families (better)

### LINKED ISSUE #716: W3 Validator error when requesting multiple google fonts family
Example:
```
Error: Bad value //fonts.googleapis.com/css?family=Fjalla+One:400|Cantarell:400 for attribute href on element link: Illegal character in query: | is not allowed.
From line 109, column 1; to line 109, column 143
='all' />↩<link rel='stylesheet' id='tc-gfonts-css'  href='//fonts.googleapis.com/css?family=Fjalla+One:400|Cantarell:400' type='text/css' media='all' />↩<link
```

Looks like we have to encode that "|" see:
http://stackoverflow.com/questions/22466913/google-fonts-url-break-html5-validation-on-w3-org

eri-trabiccolo:
fixes #716

### DIFF
--- a/inc/_dev/class-fire-resources.php
+++ b/inc/_dev/class-fire-resources.php
@@ -427,7 +427,7 @@ if ( ! class_exists( 'CZR_resources' ) ) :
 
       wp_enqueue_style(
         'tc-gfonts',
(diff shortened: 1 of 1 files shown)


eri-trabiccolo:
chose among this and #717

PR #749, opened 2017-01-22
Imp: slightly alter the breadcrumbs related code to accept pro features

eri-trabiccolo:


eri-trabiccolo:
revamped for the new customizr.
you can merge it ;)

PR #779, opened 2017-01-30
Imp: set fp dropdown pages default value to 0

### LINKED ISSUE #778: Customizer: FP dropdown pages control by default shows the latest option
Since featured pages dropdowns do not have a default value, on **fresh installs**, the js selecter instead of showing "--Select--" shows the latest 'option' as it was the selected one.

This happens even with pre-customizr-fmk versions.

eri-trabiccolo:
fixes #778


Nikeo:
merged in dev

PR #782, opened 2017-02-01
Fix: make front js correctly react to customizer partial refresh

eri-trabiccolo:


eri-trabiccolo:
Fix: fix sticky header off postMessage reaction
this looks like an old bug.

PR #789, opened 2017-02-08
Imp: improve overall plugins compatibility

### LINKED ISSUE #774: Comment form coming in product reviews
https://wordpress.org/support/topic/comment-form-coming-in-product-reviews/

What can we do for this? Limit our option just for 'normal' posts?!
But more than this, should we limit the comment form/comments lists to 'normal' posts and pages?
Or find a way to exclude post types. A multiselect in the comment options, like
"Do not display theme's comment form/comment list in.."

eri-trabiccolo:
do not render the comments template more than once

should fix #774

### DIFF
--- a/comments.php
+++ b/comments.php
@@ -13,6 +13,11 @@ if ( apply_filters( 'czr_four_do', false ) ) {
   do_action( 'czr_four_template', 'comments' );
   return;
 }
+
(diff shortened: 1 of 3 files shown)


Nikeo:
merged in dev

PR #790, opened 2017-02-08
Imp: customizer - go to pro link compliance

eri-trabiccolo:


eri-trabiccolo:
- tested free and pro on old php environment

In the welcome page, I removed the donation block, though I didn't remove the go pro section. Do we have?
https://github.com/presscustomizr/customizr/blob/v3.5.1/inc/_dev/admin/class-fire-admin_page.php#L194

Nikeo:
About the go pro in admin, as long as it is not a pop up or an annoying permanent banner, theme authors can do it from what I understand.
https://make.wordpress.org/themes/2016/01/26/discussion-advertising-in-themes/

Nikeo:
merged in dev

PR #791, opened 2017-02-08
Imp: make free theme czr4 free

eri-trabiccolo:


eri-trabiccolo:
- tested on old php environment

PR #792, opened 2017-02-08
Imp: customizer - get back the fonts/skins select on focus

eri-trabiccolo:


eri-trabiccolo:
- tested no conflicts with customizr-pro

Nikeo:
merged in dev

Issue #787, opened 2017-02-07
double social icons


denzelchia:
Hi,

User of Customizr theme and Customizr Pro Theme reporting having duplicated social icons.

Issue on WordPress.org forum
https://wordpress.org/support/topic/last-update-double-social-icon-in-footer/

Issue on helpscout
https://secure.helpscout.net/conversation/314014052/22731/?folderId=607544#thread-845085769

I think this is a bug, but I am not able to replicate it on my test site, probably because my test site is a new installation.

Thank you

Best Regards,
Denzel

eri-trabiccolo:
Hi Denzel,
the problem on helpscout is a conflicting plugin which uses a pretty old font-awesome css. In that version they don't scope the icons styling to the 'fa-[icon]' but they apply their style to all the icon-* .
We kept the icon-* for backward compatibility reasons (mainly for old users who used to custom style them).

So since now, basing on our analysis, and on users feedback the reasons for that issue are:
- server side cache (plugin, hosting, cdn) not correctly cleared 
- child-theme messy CSS (e.g. whole copy of old parent CSS in the child-theme)
- plugin conflict

We'll see if we can avoid this, but the downside will be for the old customers who are not affected by the problems above and wanted to style the social icons.

denzelchia:
Hi Rocco,

Thank you very much for these information! 👍 
Glad to know it's not a theme bug.

I will use your information as reference in support forum.

Have a nice day, :)

Best Regards,
Denzel

PR #794, opened 2017-02-08
Imp: try to avoid double social icons

eri-trabiccolo:
see #787


eri-trabiccolo:
This way we should achieve:
- no caching issue
- old fa no conflicting anymore
- user who customized the social links color will still be able to see their customization

Potential caveat:
- users who used this snippet http://presscustomizr.com/snippet/add-new-custom-socials-icons/
to add a custom social icon not present in the font-awesome set, but still ported in the options (unless they change the social link in the customizr), to see the custom icon must:
replace this `.icon-someicon:before` with `social-block .social-icon.icon-someicon:before`
(for example)

Nikeo:
OK excellent thanks. I will communicate on it in the release post.

Nikeo:
merged

PR #811, opened 2017-02-17
Fix: fix potential option loss

### LINKED ISSUE #810: Potential options loss
Looks like our system of setting a single option, merging it to the raw options (to avoid third party filtering) might cause the options loss when using Memcached.
This seems to frequently happen on certain hostings which have Memcached enabled by default.
The solution would be reverting to the wp default way to retrieve the options.

Caveats:
options then might be filtered, mostly by lang plugins which filter the options in front.

eri-trabiccolo:
fixes #810


eri-trabiccolo:
Hopefully.

Anyway I think that using wp function wp_get_all_options (as it performs an sql query if the option is not cached yet) in czr_fn_get_raw_option (https://github.com/presscustomizr/customizr/blob/v3.5.2/inc/_dev/z_fire.php#L9)
would fix the issue as well.

PR #812, opened 2017-02-19
enable accessibility for social icons 

JTS-IL:
add aria-label to the social fa icons for accessibility support


Nikeo:
Thanks merged in dev

PR #813, opened 2017-02-19
add aria-label to button - support accessibility

JTS-IL:
add simple aria-label that is the same as title to pass some automatic accessibility checks (like wave)


Nikeo:
Thanks merged in dev

PR #835, opened 2017-03-08
Imp: improve slider loader visibility handling

eri-trabiccolo:
- add js detection inline script in wp_head
(ref. twentyseventeen and below)


Nikeo:
merged in dev

Issue #844, opened 2017-03-12
czr_fn_isprevdem : potential cache issue


Nikeo:
https://wordpress.org/support/topic/website-looks-different-when-viewed-live/#post-8904262

> Hello,
> 
> I need some Help. I am using the Customizr Theme for my web site. Yesterday I notice that my web site started to change all on its on. The color of the font, pictures, page layout ect. When I log into my account my web page looks like correct. I have tried to republished the site but I am getting the same results. I log into my account and everything looks fine. If I go directly to my site onlinetruevine.org it looks completely different. Can Someone help me fix this problem?

eri-trabiccolo:
Not sure Nico,
I don't think from cache should be a param.
We're never interested in bailing if the value is not cached.
Just a question: when you return alloptions if optname is null...

What's the purpose? To return the alloptions['opt_group'] in that case?
But that would mean that opt_group is not null and or has been found. Otherwise you really might return alloptions as is.

So my question is, is there a case where you're really interested in getting $alloptions?

eri-trabiccolo:
what do you think about #846 
maybe we can remove the whole true == wp_installing() case and special case like home, siteurl..etc we're not interested in.
That's pretty much a copy of wp get_option but without filters plus "suboption" handling.

Nikeo:
Thanks for your feedbacks !
Let's discuss them on Monday. Enjoy the dauy of the Lord. ( and I know it means so much for you ... ) ;)

eri-trabiccolo:
Yes, thanks for respecting my feelings ;)

PR #846, opened 2017-03-12
Imp: get_raw_option as get_option with suppress filters

eri-trabiccolo:
a different fix version for #844 
plus prevents also the overwrite of theme options with filtered options values (see lang plugins) in front


eri-trabiccolo:
ah this is old, should not be merged

PR #852, opened 2017-03-15
Imp: display thumbnail in single pages like we do in single posts

### LINKED ISSUE #738: [Feature] Show featured image in pages too
https://secure.helpscout.net/conversation/306135677/21712/?folderId=607709

eri-trabiccolo:
fixes #738


eri-trabiccolo:
It consists in a duplication of the single post (a)code and (b)customizer options.
(a) Even if a lot of the code is shared there are still specific contextual logic. I renounced to generalize the code and move it to a "shared" place as it would have been expensive in matter of time.
(b) I considered they should have been totally separated

Nikeo:
good choices. Agreed. Not a big deal to not reach the perfect code structure for this, considering the other on going priorities.
thanks for the explanations

PR #854, opened 2017-03-16
Fix: allow using first attachment as thumb for slider of posts in pages

eri-trabiccolo:


eri-trabiccolo:
Fastest solution to me, unless we wanted to highly change the thumbnail retrieval logic...

PR #862, opened 2017-03-24
Imp: allow menu centering when logo/title centered

eri-trabiccolo:


assets/czr/_dev/js/czr-control-dom_ready.js
```diff
@@ -119,6 +119,40 @@
                           return state.text;// optgroup different than google font
                         return '<span class="tc-select2-font">' + state.text + '</span>';
                     }
+
+                    /**
+                    * Dependency between the header layout and the menu position, when the menu style is Side Menu
```
Nikeo: couldn't you use the dominus/servi code here, with an action callback ?

eri-trabiccolo: yeah I could (should).
But I wanted to fire this on ready too, and my latest tests reported that the section's awekening (as the dominus and servi are located in different sections) didn't work.
Maybe now is fixed, or I always did something wrong :D

PR #876, opened 2017-04-19
Fix: use wp_sprintf to avoid Warning sprintf(): Too few arguments

### LINKED ISSUE #875: Warning: sprintf(): Too few arguments in /wp-content/themes/customizr/inc/czr-front.php on line 6918
https://wordpress.org/support/topic/warning-sprintf-too-few-arguments-chinese-only/
> Customizer 3.5.11
> 
> I’m asking here first since the error specifies a file in the theme.
> 
> Warning: sprintf(): Too few arguments in /home/huayecss/public_html/wp-content/themes/customizr/inc/czr-front.php on line 6918
> 
> Image example
> 
> This happens only in Chinese. English and Spanish do not have this error. I created a test post in Chinese (via Polylang) and the error appears. If I simply change the language setting to English, it vanishes.
> 
> It happens if the post is entirely written in English, but set to “Chinese”, so it’s not the Chinese characters causing it. It does it with any author, so it’s not that. It does it with all meta info entered (alt text, title, etc.) and with all left empty. And it happens in any screen that shows the thumbnails (categories, author, news).
> 
> As we’re a business site with visitors 24/7, I’d prefer to not go through the “turn off all plug-ins” procedure if possible.
> 
> If anyone has any ideas as to the cause, I’d appreciate hearing them.

eri-trabiccolo:
fixes #875

### DIFF
--- a/inc/_dev/parts/class-content-post_metas.php
+++ b/inc/_dev/parts/class-content-post_metas.php
@@ -179,10 +179,17 @@ if ( ! class_exists( 'CZR_post_metas' ) ) :
           $upd_date   = $this -> czr_fn_get_meta_date( 'update' );
 
           $_args      = compact( 'cat_list' ,'tag_list', 'pub_date', 'auth', 'upd_date' );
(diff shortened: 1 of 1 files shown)


ghost:
merged in dev

PR #888, opened 2017-05-17
Imp: fix compatibility with polylang-pro

### LINKED ISSUE #887: Plugins compatibility: Polylang-Pro compatibility issue
The theme is not compatible with Polylang-Pro while it is with Polylang free version

eri-trabiccolo:
fixes #887


eri-trabiccolo:
This doesn't seem to be actually merged in dev @nikeo 
probably the front js fmk update merge erased this?

PR #889, opened 2017-05-17
Update select2

eri-trabiccolo:


eri-trabiccolo:
This doesn't seem to be actually merged in dev @nikeo 
probably the front js fmk update merge erased this?

PR #890, opened 2017-05-24
New grunt czr4

eri-trabiccolo:
TODO:
unify old and new js files like waypointm request animation frame etc etc.
The old theme should refer to the new path


eri-trabiccolo:
pr made against a wrong branch, I'm closing it

PR #891, opened 2017-05-24
New grunt czr4

eri-trabiccolo:


eri-trabiccolo:
TODO:
unify old and new js files like waypointm request animation frame etc etc.
The old theme should refer to the new path

ghost:
I'm adding the todo as a new issue.
Thanks

PR #901, opened 2017-05-26
Fix: for czr4 alpha testing always fire the customizer in dev/debug mode

eri-trabiccolo:
czr4 customizer concatenated js are not built, they are in a very early stage, let's be sure we always enqueue the dev files.


ghost:
need to implement concatenate and minify tasks

PR #909, opened 2017-05-31
Customizer unify

eri-trabiccolo:
customizer unification
contains #908 #902


eri-trabiccolo:
About the customizer unification:

My plan to have common "parts" (meaning the customizer modules, controls etc) but different czr-customize (meaning the concat of czr-init and czr-resources) changed as I realized that there were too many shared functions in czr-init (e.g. the customize_factory is there, and we don't want to have it in several places).

I preferred to merge everything in the new theme customizer with some (not so many) switches depending on whether or not we're running c4.

The alternative would have been a czr-init and czr-resources parcelling, so to keep a "common base", and then a different concatenation for the c4 and the old.

PR #916, opened 2017-06-02
New pro compat

eri-trabiccolo:
new pro compatibility
contains #908 #902 #909 and #915


eri-trabiccolo:
At the moment, though, both the pro and the free (packages) customizer will not work in czr4, as we don't copy dev files and czr4 customizer always looks for dev files. => TOFIX

PR #953, opened 2017-06-13
Imp: prepare theme for future modern style in pro

eri-trabiccolo:
do not allow theme style switching via $_GET when is Pro
while will still be possible by setting a constant in wp-config.
Since the style switch is done very early, we need to setup some
constants (like the ones about the theme name) before the style
switch.
For this reason the whole setup constant function has been moved
in the shared file functions-base.
Additionally init-pro file will not be concatenated anymore in the classic
style init (inc/czr-init.php), but will be required by init-base before
the style switch.


eri-trabiccolo:
needs a build to be effective

PR #960, opened 2017-06-15
Imp: improve deploying process to avoid headers already sent issues

eri-trabiccolo:


eri-trabiccolo:
I reproduces the issue with the 2 new lines at the end of a file we concatenate.
I made a build.
I tested it on the test env where I can reproduce the headers already sent (silent) issue in the customizer. => OK
I tested this build with theme check and no multiple line endings issue was raised.

PR #974, opened 2017-06-18
Fix: fix access to undefined tc_rectangular_size class property

### LINKED ISSUE #971: Warning: array_merge(): Argument #1 is not an array in [path]/wp-content/themes/customizr/core/init-base.php on line 694
https://presscustomizr.com/customizr-release-note-for-version-3-5-16/comment-page-1/#comment-369653

> Throwing error of Warning: array_merge(): Argument #1 is not an array in [path]/wp-content/themes/customizr/core/init-base.php on line 694

> every page on the admin, that is. At least the public pages aren’t showing it.

eri-trabiccolo:
fixes #971


eri-trabiccolo:
This is how it was in the classical class-fire-init.php

PR #1002, opened 2017-06-26
Fix: classic - fix non existing function as __ID filter callback

eri-trabiccolo:
czr_fn_id should be replaced by czr_fn_get_id.
czr_fn_id exists only as static (deprecated) method of the classic
CZR_utils class


eri-trabiccolo:
This is minor, as we don't use the ___ID filter as far as I can see.
Might be used by some users in child themes.

ghost:
fixed locally

ghost:
thanks larry

PR #1048, opened 2017-07-20
Bs job

eri-trabiccolo:


eri-trabiccolo:
I work on the concatenation on another branch

PR #1069, opened 2017-07-22
Fix: modern - fix related posts header margins ridiculously high

eri-trabiccolo:
@Nikeo 
take a look at this before releasing please


ghost (commented):
Yep 100% agree.
I would even decrease this margin top to 1em since the rel. post title already applies a significant margin top.
thanks

PR #1070, opened 2017-07-23
Rtl compliance

eri-trabiccolo:


eri-trabiccolo:
This PR, along with https://github.com/presscustomizr/customizr/pull/1068,
covers the major issues I could find with the current theme.
Further work will be needed, but if you have the time to test both, would be better to include them in the release

Issue #1103, opened 2017-07-26
Various minor problems reported in the release note


ghost:
http://presscustomizr.com/wp-admin/edit-comments.php?p=210847&comment_status=approved

> Hi Nicholas, great job on this theme, I find it awesome!
> 
> However, I found some bugs after switching to the modern style:
> – The hierachy of categories is not shown in the categories widget
> – Featured pages don’t appear on the main page
> – The word “Archives” is doubled in the archives widget
> 
> When I switch back to the classical style, the problems above are no more. Is it only my page or have you already heard about those bugs?
> 
> Keep it going, I think it’s one of the best WordPress themes!
> 
> Best regards,
> Michal

eri-trabiccolo:
1. fixed in #1105 
2. fixed (as far as I can see) in #1098
3. fixed in #1106

Issue #1102, opened 2017-07-26
List Formatting


ghost:
> Hi Nicolas,
> 
> Thanks for a great website, which I have been using for some time now. Regarding “modern”, the following questions:
> 
> **– Is it intentional that indents (bullet points and numbering) are no longer indented?**
> 
> Best regards
> Robert

eri-trabiccolo:
Cannot reproduce it.
Maybe a plugin resets the indentation?
![list_indentation](https://user-images.githubusercontent.com/7689242/28610583-120fb01c-71e9-11e7-9a9c-39bff4fa0996.png)

Anyway I push a fix to use the same indentation for ul and ol

eri-trabiccolo:
Ah I see, we're talking about the right sidebar.
Ok

eri-trabiccolo:
fixed in #1105

PR #1105, opened 2017-07-26
Fix: modern - same indentation for ul and ol

eri-trabiccolo:


eri-trabiccolo:
The second commit of this PR should fix #1102
 certainly fixes #1103 bullet 1

PR #1178, opened 2017-09-20
Gallery img sizes

### LINKED ISSUE #1165: Galleries : the thumbnail should use a larger image size in most of the case
150x150 will look pixelized in most of the cases

https://demo.presscustomizr.com/gallery/

@eri-trabiccolo any simple way to fix this ?

eri-trabiccolo:
fixes #1165

### DIFF
--- a/assets/front/scss/0_3_components/_gallery.scss
+++ b/assets/front/scss/0_3_components/_gallery.scss
@@ -48,39 +48,69 @@ section[class^="post-"] .gallery.row {
   }
 
   //
-  // Gallery needs a better implementation
+  // gallery columns management inspired by twentyseventeen
   //
-  @for $i from 1 through 9 {
-    &.gallery-columns-#{$i} {
-      .gallery-item {
-        //width: calc( (100% - #{$base-line-height} * (#{$i} - 1) ) / #{$i} );
-        width: percentage( 1 / $i );
-      }
-    }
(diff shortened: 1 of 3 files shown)


ghost:
looks good, thanks for the PR and the timing.
I'll merge it later today.
Let me know if there's anything I need to know about the technical choices / assumptions you've made. As far as I can see, you've commented a lot so I think it's fine !

eri-trabiccolo:
Hi Nico.
There are no great assumptions.
When implementing the "sizes" attribute, I just needed to make the image width for each relevant viewport matched the gallery item widths, and I realized that the CSS that defined the number of columns for a given viewport wasn't really good to me.
So first of all I changed it inspired by 2017theme.
Then I applied the same "maths" in PHP.
I tried to comment as much as possible my choices and "maths" (no rocket science of course), and keep the code more verbose at the expense of compactness just to make it more clear.

The only other thing you might need to know (but I think you already saw it :D) is that I didn't use a filter to define the sizes attribute, as it can be passed to the wp functions when retrieving an image attachment.

p.s.
The second commit (https://github.com/presscustomizr/customizr/pull/1178/commits/18b32d207ce403d9c92cd1df6b4b0b9719a006c3) is a starting point, to me, to unify some variables in php/css and js in the future (see media queries breakpoint)

PR #1187, opened 2017-09-26
Fix: modern - fix colors of the search form in dark overlay

### LINKED ISSUE #1185: Search form colors in the overlay: don't take the "skin" in account
https://github.com/presscustomizr/customizr/blob/a7088476905ea9a3a497b09e6d418c63423d5e6a/assets/front/scss/0_2_header/skins/_skins.scss#L225

One of the changes in the following PR
https://github.com/presscustomizr/customizr/pull/1134/files

doesn't seem to be correctly merged (probably because of a conflict).

The effect of this is that the full page search form, when the overlay has a dark background, isn't very visible.

eri-trabiccolo:
fixes #1185

### DIFF
--- a/assets/front/scss/0_2_header/skins/_skins.scss
+++ b/assets/front/scss/0_2_header/skins/_skins.scss
@@ -222,7 +222,7 @@ $secondary-color-dark   : $grey;
   }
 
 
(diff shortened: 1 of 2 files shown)


ghost:
ok thx

PR #1190, opened 2017-09-27
Imp: modern - improve ol/ul margins in tc-content-inner (.entry-content)

### LINKED ISSUE #1183: ul, ol margins
Following this feedback http://presscustomizr.com/release-note-for-customizr-pro-v2-0-9-and-customizr-free-v4-0-8/comment-page-1/#comment-375711

> With the option “Modern”, the theme handles bullet point and numbering differently compared with “Classic” and what is considered correct. I.e. with bullet points and numbering, the second and following lines normally indent to the start of the text in line one.
> 
> Is the intention with “Modern”, that it refrains from doing so, or will it be adjusted?

This user raise a good point, I would prefer having a left margin ( right for RTL ) on lists, just like here for example : http://docs.presscustomizr.com/article/3-creating-a-slider-with-customizr-wordpress-theme#remove-default-slider
This margin should be applied only for the lists inside a page or post content.
I think we already discussed this point a few monthes ago.

@eri-trabiccolo from your standpoint, would there be any unwanted side effect ( with plugins or smthing else) of modifying the current css rules for ul, ol, which are 
```
ol, ul {
    margin: 0;
    padding: 0;
}
```

to something like

```
.entry-content ol, .entry-content ul {
    margin: 0 0 9px 25px;
    padding: 0;
}
```

To be discussed next monday !

eri-trabiccolo:
fixes #1183

also slightly improve the cite element style

### DIFF
--- a/assets/front/scss/0_3_components/_components.scss
+++ b/assets/front/scss/0_3_components/_components.scss
@@ -451,7 +451,14 @@ blockquote, .entry-link {
     @extend .bold;
     &::before{
       top:$base-line-height*2;
-      left:$base-line-height*2;
+
+      @if ( true == $is_rtl ) {
+        right: $base-line-height*2;
+      }
+      @else {
+        left:$base-line-height*2;
+      }
+
       position: absolute;
       font-size: ms(2);
       display: inline-block;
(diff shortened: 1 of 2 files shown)


ghost:
really clean thanks

PR #1198, opened 2017-10-01
Header custom skin

### LINKED ISSUE #1193: Header style : add new option for custom color
- There should be a "Custom" option in addition to the current Light and Dark options.
- When "Custom" is selected, 2 color pickers are revealed => needs to add a new dominus / servus dependency
  - a background color picker : default color == light
  - a text color picker : default color == #313131
- For the menu _progress bar effect_ on hover in the header, the user text picked color will be treated in php to reproduce the scss invert
- The setting transport for the color pickers will still be "refresh", even if not ideal

eri-trabiccolo:
fixes #1193


eri-trabiccolo:
Hi,
give it a try.
I tested it and works fine but I had to take some decisions that I documented (via inline comments), so let me know if you don't agree.
There's a small merge conflict because I started this dev before you merged this:
https://github.com/presscustomizr/customizr/pull/1187

As you can see, it's not a big deal:
https://github.com/presscustomizr/customizr/pull/1198/conflicts

the header-custom-skin branch version should be kept there, so:
`.tc-header .czr-form, .czr-overlay .czr-form {`
is the line to keep

ghost:
Excellent. your color choice for the progress bar is good for me too.
I like the let's dance part :)
I don't think I got the entire picture about grey light and dark, but this is not a problem. The final result is really good.
Merging now. thx

ghost:
All my tests went fine.

I've tested various scenarios with regular and side menus.
top bar + socials
mobile / desktop
hover effect on submenu items

PR #1213, opened 2017-10-09
Fix: modern - fix mobile menu id duplication

eri-trabiccolo:
the mobile nav container and the ul menu elements
happened to have the same id


ghost:
good catch (thanks w3c check :) )

PR #1214, opened 2017-10-09
Imp: modern - no need for the hamburger link to have a button type

eri-trabiccolo:
attribute
plus it generates html validation error


ghost:
ah yes

PR #1276, opened 2017-10-31
Fix: modern - fix blockquote and czr-links in small device viewports

eri-trabiccolo:
additionally: in alternate, avoid the content wrapper growing its width
when a text has long words (because of the flex system in place)


ghost:
ok cool. good catch

PR #1278, opened 2017-11-01
Fix: modern - fix trying to access not defined variable in the slider

### LINKED ISSUE #1277: Notice: Undefined variable: image in core/front/models/modules/slider/class-model-slider.php on line 323
previously in https://github.com/presscustomizr/pro-bundle/issues/110

eri-trabiccolo:
module
while composing the slides srcset to add the small slider thumb size
and the original slide image edited in WordPress after the upload.

fixes #1277

### DIFF
--- a/core/front/models/modules/slider/class-model-slider.php
+++ b/core/front/models/modules/slider/class-model-slider.php
@@ -318,19 +318,15 @@ class CZR_slider_model_class extends CZR_Model {
 
               //always use the tc-slider-small if available and not corrupted
               //and that is not from previous edits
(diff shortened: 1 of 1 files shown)


ghost:
Thank you for the quick fix Rocco.
I'm merging now.
I'm polishing things for the flickity galleries and about to push live for both versions.

eri-trabiccolo:
Ok, I was looking to more performance improvements for masonry and infinite.
Not huge boost though. Anyway when you have time we can discuss.

PR #1297, opened 2017-11-10
Fix: modern - fix potential slider captions overlapping on page load

### LINKED ISSUE #1293: Main slider captions overlapping on page load
Reported here:
https://secure.helpscout.net/conversation/462551520/34125?folderId=607709

I cannot reproduce it on my end, though is visible on the user's website.
What is actually weird is that the slides should not be visible because of the loader overlapping them.
But it seems that in this case the slider loader appears with a certain delay. Most likely due to the caching/concatenating plugin in place which loads some heavy js a the very top of the page.
**Note:** The slider loader is hidden when the html tag has the class 'no-js' (default), it appears only when this class is replaced with 'js'. We perform the html class replacement here:
https://github.com/presscustomizr/customizr/blob/v4.0.11/core/init-base.php#L904

This is something that can actually happen for the way the main slider's slides are made.
Basically the `.carousel-caption`s are absolutely positioned, and they refer to the first positioned (not static) ancestor, which, when flickity is not enabled, is the `.czr-slider-holder`, not the `.carousel-cell`s.
**Note:** `.carousel-cell`s, on load, having a static position and width 100%, are displayed one of top of the other, but just the first one is visible, as the slider wrapper hides what overflows.

## Solution?
So the solution would be to set the `.czr-carousel .carousel-cell` position relative (flickity then, once enabled, will set the position to absolute).

This way, the cells will still stay one on top of the other, but each caption position will refer to the cell, hence only the first caption will be visible, as it should.

eri-trabiccolo:
fixes #1293

### DIFF
--- a/assets/front/scss/0_3_components/_components.scss
+++ b/assets/front/scss/0_3_components/_components.scss
@@ -865,6 +865,12 @@ html.no-js [id*=czr-slider-loader-wrapper] { display: none; }
   position: relative;
   overflow: hidden;
 
+  .carousel-cell {
+    position: relative;
+    width: 100%;
+    overflow: hidden;
+  }
+
   @at-root .carousel-caption {
     height: auto;
     max-height: 80%;
--- a/assets/front/scss/0_3_components/_temp.scss
+++ b/assets/front/scss/0_3_components/_temp.scss
@@ -6,15 +6,7 @@
 }
 
 
-
-/* FLICKITY SLIDER  */
-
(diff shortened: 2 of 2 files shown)


ghost:
cool, that was simple. Are you sure it plays well with the gallery carousel ?
To be discussed now

eri-trabiccolo:
well, yes, tested without issue. Consider that in the galleries carousels the cells contain only the image, and as soon as the flickity is initialized the cells become absolute.

Issue #1204, opened 2017-10-06
Feature request: submenu collapsing in mobile and sidenav menu.


eri-trabiccolo:
Since the mobile menu now is a totally different menu we can add an option to display the submenu collapsed (uncollapsing on click) as it won't interfere with the "horizontal" menus.
Same can be done for the sidenav menu.

With some additional work we can then port this for the classical style (meaning porting also the different mobile menu there).

ghost:
@eri-trabiccolo to be discussed during the next point

ghost:
Ah yep
cool !

ghost:
> I like the new modern style of your theme. I’m working with the pro version and actually starting to built my new site. My database crashed and so I started to built a completely new site. I’m not so happy with the style of the mobile menue. Is there a possibility to change the fact, that all items are open? Think about click on the parent items to open the childs? I have used Max Mega Menue for my previous site. The mobile menue looked nice but I would like to built my new site without the plugin. Do you have some advices how to manage the mobile menue more comfortable? Thanks a lot.
> Birgit
> P.S. Because I could not see my comment on my PC I post it again. Sorry, when it is a second comment now.

PR #1301, opened 2017-11-15
Vertical menu dropdown

eri-trabiccolo:
php/css part for https://github.com/presscustomizr/customizr/issues/1204
needs https://github.com/presscustomizr/customizr-front-js-parts/pull/23

Note:
- In vertical navs, when the dropdown on click is enabled: to better handle the alignment (and flexibility) of the link + caret (dropdown toggler) I wrap them in a span. We should consider to extend this to all our header menus, just for a matter of consistency.
- I add a class to the nav menu item links, and to the span tag that wraps the nav item title, so to have a better control over the elements, avoiding stuff like span:first-type-of in CSS for example.
- In vertical navs, when the dropdown on click is enabled: I visually highlight the "caret" dropdown toggler by setting it's background color which is the same of the regular submenu items on hover (the progress bar effect).

### DIFF
--- a/assets/front/js/libs/outline.js
+++ b/assets/front/js/libs/outline.js
@@ -27,7 +27,7 @@ var tcOutline;
 
   // Using mousedown instead of mouseover, so that previously focused elements don't lose focus ring on mouse move
   add_event_listener('mousedown', function(){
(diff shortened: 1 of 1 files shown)


ghost:

PR #1314, opened 2017-11-22
Fix: modern - move comment form before comment list (if any)

### LINKED ISSUE #1306: Comment form location => new option to change the location ?
> Can you confirm that the comment form is still not directly after content when using modern style ? I mean that the comment form is at the bottom of the pages after ALL comments. This is what I had seen, but since, I haven’t tried again.

eri-trabiccolo:
replicates the exactly classical style order
fixes #1306


ghost:

PR #1316, opened 2017-11-22
Fix: modern - fix related posts height after the content with two sid…

### LINKED ISSUE #1304: Related posts right after the content with two sidebars should have free height
It's like this at the moment:
wrong

Should be like this:
good

The easiest way to achieve that is by scoping some CSS rules:
https://github.com/presscustomizr/customizr/blob/v4.0.12/assets/front/scss/0_4_layout/_grid.scss#L953
should be 
`#content:not(.col-md-6) & section.entry-content__holder`
as well as this
https://github.com/presscustomizr/customizr/blob/v4.0.12/assets/front/scss/0_4_layout/_grid.scss#L959
`#content:not(.col-md-6) &  .entry-content__wrapper `

eri-trabiccolo:
…ebars

fixes #1304

### DIFF
--- a/assets/front/scss/0_4_layout/_grid.scss
+++ b/assets/front/scss/0_4_layout/_grid.scss
@@ -950,11 +950,10 @@ a.czr-format-link {
           padding: 0;
       }
 
-      section.entry-content__holder {
-          &::before {
-            display: block;
-          }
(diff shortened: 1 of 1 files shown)


ghost:

PR #1338, opened 2017-12-20
Fix: modern - fix mc effects control visibility in the customizer

eri-trabiccolo:
depending on the site layout: when boxed the effect is forced to
"reveal" and the effects control, therefore, must be hidden


ghost:
ok thx.
merging

PR #1341, opened 2017-12-21
Fix: modern - fix Terms and Condition checkbox in Checkout page is not

### LINKED ISSUE #1340: WooCommerce: Terms and Condition checkbox in Checkout page is not really visible
reported here on HelpScout, but the user is running the Customizr (non Pro) theme: 
https://secure.helpscout.net/conversation/491171725/35742?folderId=607541)

eri-trabiccolo:
really visible

fixes #1340

### DIFF
--- a/assets/front/scss/0_3_components/_plugins.scss
+++ b/assets/front/scss/0_3_components/_plugins.scss
@@ -40,7 +40,7 @@
     color: inherit;
   }
   #ship-to-different-address-checkbox {
-    margin: -1px;
+    margin: -3px 0 0;
   }
 
   input#submit[class*=alt],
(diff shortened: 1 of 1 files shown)


ghost:
that was fast

eri-trabiccolo:
That's a trick, I wrote the issue when I had the solution :P

ghost:
ah I see  :)
but kinda scaring...how many unsolved issues have never been opened then ? 🤐

eri-trabiccolo:
ahah I always solve them! :P 
Nope, it depends, when I see it's important I try to find a solution before writing the issue - and if I don't find it in a reasonable time i write the issue anyways-, otherwise I only write the issue.

ghost:
I knew it. You always find a solution.
The good thing with us developers, is that we define what is a "reasonable time" !  ( I totally include myself ). :D

Issue #1373, opened 2018-01-10
Mobile hamburger menu gets expanded then collapse automatically


ghost:
The bug can be easily reproduced on https://cb-electronics.com with chrome

> For the website design on mobile phones, the menu button (The three lines) when pressed, the menu doesn't appear, and it gives the sign of X instead the three lines.
> 
> Device that we tried it on: iPhone 5S - iPhone 6S+ and iPhone 7
> 
> Browser: SAFARI

> SITE_URL: https://cb-electronics.com
> HOME_URL: https://cb-electronics.com
> IS MULTISITE : No
> 
> THEME | VERSION : customizr | v4.0.14# WP VERSION : 4.9.1
> PERMALINK STRUCTURE : /%postname%/
> 
> ACTIVE PLUGINS :
> Contact Widgets: 1.4.1
> Custom Product Tabs for WooCommerce: 1.6.3
> Download Manager: 2.9.61
> GoDaddy Email Marketing: 1.2.0
> Limit Login Attempts: 1.7.1
> TablePress: 1.9
> WooCommerce: 3.2.6
> 
> PHP Version: 7.1.7
> MySQL Version: 5.6.36-cll-lve
> Web Server Info: Apache
> 
> WordPress Memory Limit: 40MB
> PHP Memory Limit: 256M
> PHP Upload Max Size: 2M
> PHP Post Max Size: 8M
> PHP Upload Max Filesize: 2M
> PHP Time Limit: 30
> PHP Max Input Vars: 1000
> PHP Arg Separator: &
> PHP Allow URL File Open: Yes
> WP_DEBUG: Disabled
> 
> Show On Front: nothing
> Page On Front: (#0)
> Page For Posts: (#0)

ghost:
@eri-trabiccolo can you investigate it in priority during your next work session ?
thx!

ghost:
I've asked the user to test without plugins. ( table press might be conflicting ). I'll update the issue once he replies.

eri-trabiccolo:
Download manager uses bootstrap3 : collapse conflict.
We should use a custom code

Il 10 gen 2018 9:39 PM, "Nicolas Guillaume (@nikeo on wp.org)" <
notifications@github.com> ha scritto:

eri-trabiccolo:
see https://github.com/presscustomizr/customizr/issues/1144

ghost:
update

> Hello Nicolas,
> Here's what I did. I deactivated each plugin alone and tested the site.
[...]

PR #1377, opened 2018-01-12
Imp: avoid mobile menu and comment tabs conflicts with bs based plugins

eri-trabiccolo:
part of the fix for #1373


ghost:
thx
now the js I'm merging the fmk js part.

PR #1459, opened 2018-02-12
Fix: modern - fix tagline visibility in mobiles when shown in desktop

### LINKED ISSUE #1458: When the tagline location is set to the topbar in desktop, it is not displayed anymore in mobile
Reported here : https://secure.helpscout.net/conversation/522310678/37742/?folderId=607544

The problem only occurs when the tagline location is set to topbar in desktops.

eri-trabiccolo:
topbar

fixes #1458


ghost:
tnx, that was fast .... !

Issue #1176, opened 2017-09-18
Bootstrap 4 released => let's update our scss + templates


ghost:

eri-trabiccolo:
From a preliminary look at the diff there are few things changed.
One of them is minor, and we were aware of: the drop for some old browsers (-webkit), that will result in a smaller CSS size (less prefixed properties).
But there are also some class changes that I need to check.

I cannot do this right now but I would be able to do it by the end of tomorrow.

ghost:
ok. Let me know what you find.
I'm planning to release a version of the free and pro theme today. This will be part of the next one

eri-trabiccolo:
Yeah, it might take a while as they totally changed the gruntfile.js and package.json
So I have to work on a way to build our min dist

ghost:
OK. So if you don't mind, let's postpone the bootstrap update to next month.
Instead, let's focus on those two short term devs / docs

- https://github.com/presscustomizr/customizr/issues/1140
- https://github.com/presscustomizr/pro-bundle/issues/91

Let me know if this can fit in your planning ?

Nikeo:
mm wait, I think that the header is too big

eri-trabiccolo:
ok,
but remember that I won't be available tomorrow.

ghost:
ok, this can be done by next Monday / Tuesday.
Not a problem.

ghost:
Let's postpone those compat check when Bootstrap offical V4 will be released. Hopefully in early 2018.

ghost:
v4.0 is officialy released
https://blog.getbootstrap.com/
Let's update our code

@eri-trabiccolo I propose that you make a preliminary analysis of what needs to be done and then we validate the specs together this week when you're ready.

eri-trabiccolo:
Custom version build:
https://github.com/presscustomizr/bootstrap-light/tree/customizr-bv4

PRs:
customizr: https://github.com/presscustomizr/customizr/pull/1467
pro-bundle: https://github.com/presscustomizr/pro-bundle/pull/142

eri-trabiccolo:
## Major changes:
- B4 drops all `transition` property prefixes for pretty old browsers.
- since we use a custom classes/js for the `collapse` we don't include anymore the b4 transitions scss part -> treated in https://github.com/presscustomizr/customizr/pull/1467/files#diff-1f9cba6b3b49542d00f2d50f826bb87bR186
[...]

PR #1467, opened 2018-02-16
Bootstrap4stable update

eri-trabiccolo:
fix for #1176


core/front/models/class-model-header.php
```diff
@@ -165,18 +165,36 @@ function czr_fn_extend_params( $model = array() ) {
...
+      (sm up), apply the following classes to your element:
+      a) d-none
+      b) d-lg-(flex|block|inline-block...)*
+      *meaning that you need to exactly know how to display it in lg, either flex
+      or block, etc.
+
+      - say you want to display something in mobiles (tablets and spartphones)
+      but not on large devices (desktop) you just need to apply this class to your element
+      d-lg-none
+    */
```
ghost, line 185: ok thanks for the very clear explanations.
Good improvement in B4 I think

ghost:
awesome work Rocco  !!

core/front/models/class-model-header.php
eri-trabiccolo: yes, not really noticeable if not in edge cases, but I think is a technically good improvement.
Basically they did what we discussed, and somewhat planned to do, when we talked about the "half pixel" issue about one year ago. It's good to see that we ended up with the same solution (even if we didn't implemented at that time :D).

PR #1481, opened 2018-02-28
Imp: modern - add horizontal widget area before the footer

### LINKED ISSUE #1369: Footer : add a widget zone before the footer, not styled like the footer => white background
It could be used to display bottom contact informations for example. Of anything a user want to add there.

Requested here  ( not exactly this way but still, same concept ) https://secure.helpscout.net/conversation/500718661/36316?folderId=607541
> I would like to change the background color of the "credits" footer (other than light or dark). My attempts to do so with CSS have failed, though I did manage to change the font color and the top border. Suggestions?
 **Also, can I add a text widget inside there, to the left of the social icon bar to include my client's contact information. Many thanks in advance,**

connected to https://github.com/presscustomizr/pro-bundle/issues/121

eri-trabiccolo:
fixes #1369

### DIFF
--- a/assets/front/css/style-front-placeholders.css
+++ b/assets/front/css/style-front-placeholders.css
@@ -38,6 +38,7 @@
 .tc-post-list-smarload-help {
     margin-bottom: 2em;
 }
+.horizontal-footer.tc-placeholder-wrap .tc-placeholder-content,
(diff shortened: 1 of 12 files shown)


ghost:
thanks, that was fast :)

PR #1493, opened 2018-03-12
Imp: improve woocommerce single product tabs style

eri-trabiccolo:
This way they would look like our comments tabs


ghost:
Cool

PR #1506, opened 2018-03-22
Fix: refer to correct front localized var in czr preview js

eri-trabiccolo:


ghost:
oh yes that was a big one !!

PR #1514, opened 2018-04-12
Fix: do not enqueue front placeholders assets when user not logged in

### LINKED ISSUE #1513: Front placeholder assets must not be enqueued when user not logged in

eri-trabiccolo:
and with proper caps
unless is CZR_DEV == true
fixes #1513


ghost:
ok thanks

PR #1516, opened 2018-04-13
Fix: modern - make sure mfp gallery left arrow is correctly aligned

### LINKED ISSUE #1515: MFP gallery left arrow not centered
This is due to the 'smart loading' of the mfp assets.
They used to be loaded before our main style with our rules overriding the mfp ones, but, now, they are smart loaded after our main style, so that mfp rules override our ones.
misaligned-arrow

eri-trabiccolo:
fixes #1515


ghost:
thanks !

PR #1517, opened 2018-04-13
Fix: modern - set word-wrap break-word for whatever text in #tc-page-…

eri-trabiccolo:
…wrap

This will avoid text overflowing the parent element width.
e.g. list items


ghost:
thx. how did you catch this one ?

eri-trabiccolo:
Just testing something else.
I've put a long <li> text without spaces and found it didn't wrap.

ghost:
ok !

PR #1522, opened 2018-04-20
Imp: modern - decrease space above WooCommerce page title

### LINKED ISSUE #1521: Decrease space between breadcrumbs and WooCommerce products page title
The WooCommerce products page title distance from the breadcrumbs (or the header if no breadcrumbs) should be the same we can see from the breadcrumbs and the standard page title

Standard page:
standard-space

WooCommerce products page (shop/product category/etc.):
products-page

eri-trabiccolo:
fixes #1521


ghost:
thanks 👍

PR #1537, opened 2018-05-30
Imp: build/delete retina version of pdf thumbnails

### LINKED ISSUE #1387: The retina php code creates also unwanted (and broken) PDF files retina versions
If you upload A.pdf, it creates 'retina' versions like:
A-{thumb-size}@2x.pdf
A-{thumb-size-2}@2x.pdf
A-{thumb-size-3}@2x.pdf

Consider that WP since 4.7 creates images and thumbs from the pdf like:
A-pdf.jpg
A-pdf-{thumb-size}.jpg
A-pdf-{thumb-size-2}.jpg
A-pdf-{thumb-size-3}.jpg

So we should rather have something like:
A-pdf@2x.jpg
A-pdf-{thumb-size}@2x.jpg
A-pdf-{thumb-size-2}@2x.jpg
A-pdf-{thumb-size-3}@2x.jpg

https://secure.helpscout.net/conversation/505992657/36606?folderId=607709

eri-trabiccolo:
Also improve remove retina images logic
fixes #1387

### DIFF
--- a/core/init-base.php
+++ b/core/init-base.php
@@ -63,7 +63,7 @@ if ( ! class_exists( 'CZR_BASE' ) ) :
 
             //add retina support for high resolution devices
             add_filter( 'wp_generate_attachment_metadata'        , array( $this , 'czr_fn_add_retina_support') , 10 , 2 );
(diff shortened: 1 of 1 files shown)


ghost:
thanks, really good

PR #1564, opened 2018-06-29
Fix: modern - make sure to respect the threaded comments wp setting

### LINKED ISSUE #1563: Threading comments seems to be allowed even if disabled in WordPress discussion settings
This should affect the modern style only.

Will look more in depth into this asap.

reported here: https://wordpress.org/support/topic/theme-allows-comment-replies-even-when-threading-turned-off/

eri-trabiccolo:
fixes #1563


ghost:
ok thanks

PR #1590, opened 2018-08-22
Fix: learnpress - avoid infinite loop when user tries to enroll a course

### LINKED ISSUE #1589: Compatibility with learnpress
> First I want to thank for such an excellent template, I like it a lot. I'm using the free template, but I found a problem with the Learnpress plugin. Users can not register for courses because the template jumps to an infinite cycle. I would like to know if I buy CUSTOMIZR PRO this error will be solved.
> 
> Thank you very much.

related to #904 ?

eri-trabiccolo:
fixes #1589


ghost:
merged

PR #1604, opened 2018-10-15
Add support for GitHub Updater

Eddy-Barraud:
Just adding the "GitHub Theme URI: https://github.com//presscustomizr/customizr/" line to enable automatic updates!
Using this https://github.com/afragen/github-updater


eri-trabiccolo:
Hi there,
thanks for this.

I'm not sure what it would add though for this specific theme.
Every time a new release is ready to be shipped the theme will be updated on the wp.org repository...

Could you please provide us with a use case?

Issue #1601, opened 2018-10-07
support for Gutenberg : cover image and other blocks


ghost:
- The Ultimate Guide to Gutenberg Image Alignments in WordPress Themes — Rich Tabor https://richtabor.com/gutenberg-image-alignments/
- Get Your WordPress Theme Gutenberg-ready — Ellen Bauer https://www.elmastudio.de/en/get-your-wordpress-theme-gutenberg-ready/
- Getting Your Theme Ready for Gutenberg — Bill Erickson https://www.billerickson.net/getting-your-theme-ready-for-gutenberg/

ghost:
will be done here https://github.com/presscustomizr/hueman/issues/702

eri-trabiccolo:
see also #1611 for blockquotes and pullquotes

eri-trabiccolo:
reopening because of https://github.com/presscustomizr/hueman/issues/702#issuecomment-439007412

eri-trabiccolo:
According to https://richtabor.com/gutenberg-image-alignments/

these are the blocks that support new content alignments:

- [x] Pullquote Block
- [x] Gallery Block
- [x] Image Block
- [x] Cover Image Block
- [x] Video/Audio Block
- [x] Table Block
- [x] Columns Block
- [ ] Text Columns Block
- [x] Categories Block (fullwidth only)
- [x] Embed Block

- [x] Media & Text

I'm going to see which and how to implement their alignment.

Edit:
As of now 15/11/2018:
- no alignwide alignfull classes are added to the Pullquote Block (probably a Gutenberg bug), anyway I "emulated" the test by manually adding those classes.
- couldn't find the Text Columns block, I guess it has been replaced by the Media & Text, that I've added to the list.
- The Table block has been aligned with some additional JS at the moment in the cases:
a) alignwide
b) alignfull and alignwide in a boxed layout

ghost:
OK Thank you.
Let's schedule a chat about this and select2 at around 2PM, what do you think ?

eri-trabiccolo:
Fine for me, talk you later then!

ghost:
https://weblines.com.au/gutenberg-blocks-wide-alignment-full-width/

What Block Types Can Be Full or Wide-Aligned?


> Assuming that your theme supports them, the new alignment options are offered for these blocks:-
> 
> - Cover Image Block
> - Image Block
> - Gallery Block
> - Pullquote Block
> - Video Block
> - Table Block
> - Columns Block
[...]

PR #1608, opened 2018-10-24
Imp: gutenberg cover image alignwide/alignfull compatibility

eri-trabiccolo:
CSS part for #1601

### DIFF
--- a/core/class-fire-resources_styles.php
+++ b/core/class-fire-resources_styles.php
@@ -686,7 +686,7 @@ if ( ! class_exists( 'CZR_resources_styles' ) ) :
                //Add some rules from sm up
                //add padding
                $css_container[]        = sprintf( '@media (min-width: %1$spx){ %2$s{ padding-right: %3$spx; padding-left:  %3$spx; } }',
-                           $css_container_widths[ 'sm' ],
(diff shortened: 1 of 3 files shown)


core/class-fire-resources_styles.php
```diff
@@ -686,7 +686,7 @@ function czr_fn_maybe_write_boxed_layout_inline_css( $_css ) {
                //Add some rules from sm up
                //add padding
                $css_container[]        = sprintf( '@media (min-width: %1$spx){ %2$s{ padding-right: %3$spx; padding-left:  %3$spx; } }',
-                           $css_container_widths[ 'sm' ],
+                           $css_mq_breakpoints[ 'sm' ],
```
ghost, line 689: @eri-trabiccolo  Was that change related to a bug reported for something else ?

eri-trabiccolo: something I noticed, that was wrong

eri-trabiccolo: the previous code generated a media query breakpoint too small, as the container width is smaller than the mq_breakpoint value.

ghost: ok ! thx

PR #1610, opened 2018-10-24
Fix: fix buddypress member activity filter issue

### LINKED ISSUE #1606: Buddy Press profile select filter doesn't work
This seems to be due to the fact that the Buddy Press select element wrapper has the too much generic `filter` CSS class, the same too much generic CSS class we use to apply the slider overlays:
https://github.com/presscustomizr/customizr/blob/dev/assets/front/css/style.css#L2966

The solution would be to "override" that for the buddypress filter, or, better, prefix our `filter` CSS class, to avoid compatibility problems in the future.

Basically that "before" overlaps the select so that it becomes not "not clickable".
Other minor CSS improvements can be done to that select appearance.

see: https://github.com/presscustomizr/customizr/issues/1532#issuecomment-430184314

eri-trabiccolo:
fixes #1606

### DIFF
--- a/assets/front/scss/0_1_base/_utilities.scss
+++ b/assets/front/scss/0_1_base/_utilities.scss
@@ -62,7 +62,7 @@ img.v-centered {
   max-width: none!important;
 }
 
-.filter {
+.czr-filter {
   position: relative;
   height: 100%;
   &:before {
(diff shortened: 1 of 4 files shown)


ghost:
thanks

Issue #1642, opened 2018-12-11
Gutenberg : style the editor container like Twenty nineteen does


ghost:

giorgioriccardi:
in case this helps: [https://wordpress.org/gutenberg/handbook/designers-developers/developers/themes/theme-support/#editor-styles](https://wordpress.org/gutenberg/handbook/designers-developers/developers/themes/theme-support/#editor-styles)

eri-trabiccolo:
pushed a PR with a "prototype" as we discussed.

Issue #1633, opened 2018-11-27
Elementor can't use the Customizr logo => let's implement the WP logo upload


ghost:
> Hi,
> 
> As shown in this recording, the logo in the Elementor header appears on other themes.
> https://screencast-o-matic.com/watch/cFXO28rFwD
> 
> There is a conflict with Customizr that you cannot resolve, so please refund the $89.
> 
> Thanks,
> 
> Bill

Old logo option will be used if exists
in the customizer, only the WordPress logo upload setting will be available

=> this means that we drop support for svg and gif, for new logo upload. But this will still be allowed if old users don't update logo.

ghost:
@eri-trabiccolo no rush on this. We'll discuss it

eri-trabiccolo:
In our chat we decided to:
1) display the **current** logo option choice if set, and no wp logo set, in front 
2) remove the **current** customizer logo control
3) display only the **wp** logo control

This, while enough in most of the cases, doesn't cover the case when an old user having a logo set decides to not use it anymore.

So if we want to stick on what we decided, we can propose, as FAQ for example, an hack to reset the old logo which would be:
"Set a new logo and then remove it, then publish your changes"

Then we just need a way to differentiate a wp logo never set from a wp logo not set (matter of defaults).

ghost:
Yes I'm OK with this solution : we clear the old logo option when the new wp logo is set. This way, setting a new logo and removing it could do the trick. 
The clearing of the old logo could be done on `do_action( "customize_save_{$id_base}", $this );`
see https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-customize-setting.php#L522

eri-trabiccolo:
Ah yeah makes sense, thanks for the tip!

eri-trabiccolo:
I'm not sure clearing the old logo is something we should do... you know I'm always frightened when is about manually remove theme options :) 

Anyway a PR is on the way.
Consider that removing the old logo uploader we also remove the customizer LOGO "subsection" title, and the logo control itself will sit at the very top of the Site Identity section. We might want to improve the ordering of that section with js?!

eri-trabiccolo:
Also, the SVG upload is still allowed, but the cropping need to be skipped.

ghost:
@eri-trabiccolo I'm releasing Customizr now but there's an issue with the new WP logo.
How to reproduce :

1. Set a logo on the previous version of customizr
2. Update Customizr

[...]

PR #1646, opened 2018-12-17
Fix: fix old customizer logo not shown in live preview even if not ov…

eri-trabiccolo:
…erridden

see https://github.com/presscustomizr/customizr/issues/1633#issuecomment-447592721


ghost:
thanks for the fix !

ghost:
I'm waiting for the fix https://github.com/presscustomizr/customizr/issues/1642  to release it

eri-trabiccolo:
Ah!
Ok will accelerate making it simple (meaning that I will barely copy the 2019 one) then: the thing is that I'm not very convinced about that, mostly because of possible plugins conflicts - which one? I don't know, but I know that they can occur because the block editor html/CSS is not done in a way that just changing a param (say the content width) everything will be scaled accordingly.

ghost:
OK let's do like that.
I propose that we postpone the 14:30 chat until you have finished a first prototype that we can discuss. Fine for you ?
( no problem if this is tomorrow ! )

PR #1660, opened 2019-01-10
Imp: modern - add an option to remove the various transparent header …

### LINKED ISSUE #1624: Add an option to remove the various header borders/box-shadow
It could be nice especially when combined with the new header transparent option.

requested here:
https://secure.helpscout.net/conversation/704987161/166052?folderId=607709

When the header transparent option is on we remove the box-shadows by default but we keep the tiny borders below the topbar and the header.

@Nikeo when you can let me know if we can consider to accept this as feature request.

eri-trabiccolo:
…borders

fixes #1624


ghost:
simple and right to the point. thx

PR #1663, opened 2019-01-13
Update related_posts.php

sillyonthestreet:
Useful for a fixed widget plugin, letting the plugin know where the widget should stop floating.

https://wordpress.org/support/topic/how-to-add-an-id-to-an-element/

### DIFF
--- a/templates/parts/modules/related-posts/related_posts.php
+++ b/templates/parts/modules/related-posts/related_posts.php
@@ -4,7 +4,7 @@
 */
 ?>
 <?php if ( have_posts() ) : ?>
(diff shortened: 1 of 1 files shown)


ghost:
Thanks for this proposition. I usually don't accept too specific requests for a particular plugin. But in this case, this is an harmless modification. So I'll go for it.

ghost:
merged in dev

sillyonthestreet:
Maybe I'm a bit late, but should all elements near the footer have an ID? Like the author box and comments box. Some people may not use the related posts option. 

Just a thought. Up to you.

PR #1673, opened 2019-01-25
Imp: extend anchor smooth scroll range of action

### LINKED ISSUE #1662: smooth scroll only works inside the content wrapper ?
> It seems that the smooth control only works from main content to main content.
> 
> If the Link is in the main navigation or for example in the slider, the smooth scroll is broken. 😊
> 
> # SITE_URL: https://aihro.de

Is that true. What would be the problem to listen to click events on selectors outside main content, like in header and footer ?

eri-trabiccolo:
fixes #1662


ghost:
thanks

PR #1681, opened 2019-02-06
Fix: avoid Chrome visible line issue by removing the perspective rule…

eri-trabiccolo:
… where not necessary

see https://github.com/presscustomizr/tc-unlimited-featured-pages/issues/133

### DIFF
--- a/assets/front/scss/0_1_base/_utilities.scss
+++ b/assets/front/scss/0_1_base/_utilities.scss
@@ -143,7 +143,6 @@ img.v-centered {
   margin: 8px auto;
   position: relative;
   @include backface-visibility( hidden );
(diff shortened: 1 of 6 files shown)


ghost:
related ? https://wordpress.org/support/topic/featured-image-are-surrounded-by-a-line/

eri-trabiccolo:
Yes totally related. Also the guy with the custom CSS is the pro user who reported it in the first place.

ghost:
unbelievable number of years we needed to work on this, and commits, to polish this feature / effect !

eri-trabiccolo:
yes, but I think this latest issue has been introduced recently by chrome...

PR #1685, opened 2019-02-14
Fix: give placehoder images width and height attributes

### LINKED ISSUE #1684: GT metrix says placeholder images miss width and height attributes + double slash in url path
I've also noticed that the url of the image has 2 slashes

> 
> Hello,
> 
> GTMetrix says
> The following image(s) are missing width and/or height attributes.
> 
> https://cockie.de/wordpress/wp-content/themes/customizr/assets//front/img/thumb-standard-empty.png (Dimensions: 300 x 300)
> 
> How or where can i change this?
> 
> The page I need help with: https://cockie.de
> 

@eri-trabiccolo is there a quick patch for that ?

eri-trabiccolo:
plus fix double slashes in their URL.

fixes #1684

### DIFF
--- a/core/_dev/_utils/fn-3-thumbnails.php
+++ b/core/_dev/_utils/fn-3-thumbnails.php
@@ -302,10 +302,10 @@ if ( ! function_exists( 'czr_fn_get_placeholder_thumb' ) ) {
       $_requested_size = 'thumb-medium';
 
     //default $img_src
-    $_img_src = czr_fn_get_theme_file_url( CZR_ASSETS_PREFIX . "/front/img/{$_requested_size}.png" );
+    $_img_src = czr_fn_get_theme_file_url( CZR_ASSETS_PREFIX . "front/img/{$_requested_size}.png" );
     if ( apply_filters( 'czr-use-svg-thumb-placeholder', true ) ) {
         $_size = $_requested_size . '-empty';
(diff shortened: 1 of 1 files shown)


core/_dev/_utils/fn-3-thumbnails.php
```diff
@@ -323,25 +323,37 @@ function czr_fn_get_placeholder_thumb( $_requested_size = 'thumb-standard' ) {
...
+      'thumb-standard' => array( '300', '300' )
+    );
 
     $attr = array(
       'class'             => 'czr-img-placeholder',
       'src'               => $_img_src,
       'alt'               => trim( strip_tags( get_the_title() ) ),
-      'data-czr-post-id'  => $_unique_id,   
+      'data-czr-post-id'  => $_unique_id,
+      //$_requested_size is always forced to have a value present in this array $_sizes = array( 'thumb-medium', 'thumb-small', 'thumb-standard' );
```
ghost, line 340: OK, yes I see how this is done earlier in the code. no need to check for entry existence then.

ghost:
Quick question before merging : can you confirm that setting width and height has no unwanted effects on the placeholder images in all types of grids ? For example if the centering is disabled, or other things ?

eri-trabiccolo:
Couldn't spot any issue.
We use that function only for the standard grid and the related posts (both in modern style).
Also width/height attributes are the real image ones :D

ghost:
yes, just to be sure ! :)

PR #1707, opened 2019-03-18
Fix: fix compatibility issue with the Event Tickets plugin

### LINKED ISSUE #1700: Event Tickets: Attendees Registration page shown as list of posts (when shown)
(when shown) means that if you're running the parent theme, because of a bug in the plugin code, this page is not shown at all => blank page.

But once "fixed" this bug... an archive page would be shown in place of a singular page.

reported : https://secure.helpscout.net/conversation/790525446/174966?folderId=607709

eri-trabiccolo:
fixes #1700
plus minor CSS compatibility improvement


ghost:
thanks

PR #1720, opened 2019-03-26
Fix: modern - add a new option to control the current menu item highl…

### LINKED ISSUE #1718: Menus => disable by default  the styling ( underline, background, ... ) of contextually active item
- add a new check option in header > navigation menus. disabled by default. even for old users
- option title : Highlight contextually active menu items

Important when testing => make sure it is disabled for all contextually active candidates : ancestor, parent, etc

eri-trabiccolo:
…ighting

fixes #1718


ghost:
Tested. Works fine on my end. thanks

PR #1774, opened 2019-08-29
Fix: fix modern post format meta boxes for the block editor

### LINKED ISSUE #1773: Post format metaboxes not working in the block editor if the editor loads with the Status & Visibility accordion collapsed
When the block editor loads with the accordion below collapsed
Schermata 2019-08-29 alle 12 16 07

the post format selector is not rendered yet hence our js code that bases on its value to show/hide the metaboxes doesn't init.

When this happens the only way to make it work again is:
1) uncollapse that accordion
2) reload the page so that the block editor, which "caches the preferences", will load with that accordion uncollapsed.

A fix is on the way.

n.b. of course this affects hueman too.

eri-trabiccolo:
fix #1773


ghost:
OK Thank you for that.
And thanks for cleaning the code too :)

PR #1793, opened 2019-12-16
[modern] Force categories to be shown below the title in plain text f…

### LINKED ISSUE #1792: Consider category location when using grid with plain full layout
When using plain full / excerpt layout for the post grid, when there are categories for a post they are displayed in a dedicated column on the left of the title + content/excerpt.

It looks nice when the post is short, but for long posts, it takes too much space.

As this user reports it here : https://secure.helpscout.net/conversation/1023473250/219501?folderId=607547, : 

> . For posts which have categories assigned, the categories are shown aside to the post. Which takes a significant amount from the post space. The customizing options state however that the "post meta" information is shown "below the post title", which is not correct in this case. Disabling the categories helps, however then you also don't have any categories anymore, which are there for a reason.

For plain full layout, let's display categories on top of the post title, like they are for the alternate layout.

eri-trabiccolo:
…ull content post lists layout

fixes #1792

### DIFF
--- a/core/front/models/content/post-lists/class-model-post_list_plain.php
+++ b/core/front/models/content/post-lists/class-model-post_list_plain.php
@@ -32,11 +32,14 @@ class CZR_post_list_plain_model_class extends CZR_Model {
   * return model params array()
   */
   function czr_fn_extend_params( $model = array() ) {
-      //merge with args
+      // merge with args.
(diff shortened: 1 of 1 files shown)


ghost:
Thanks for cleaning the code :)

